PR #6, opened 2021-07-20
Add USWDS to app

Sgtpluck:


app/javascript/packs/application.js
```diff
@@ -11,3 +11,6 @@ import "channels"
 Rails.start()
 Turbolinks.start()
 ActiveStorage.start()
+
+import 'uswds/dist/css/uswds.min.css'
+import 'uswds/dist/js/uswds.min.js'
```
jduss4, line 16: I don't think we've talked yet about style guide preferences, etc, but I typically have come from a school of thought where all files end with a return line.  What are your thoughts on that?

Sgtpluck: indifferent, so happy to implement it! let me see if there's a way to get out linter to yell at me about it too.

Sgtpluck: (the answer for the linter is no, not really)

Sgtpluck:
@jduss4 added the newline

jduss4 (approved):

PR #12, opened 2021-07-22
Get app into deployable state for cloud.gov

Sgtpluck:
This PR has two pieces
1. Changing the way we're pulling in USWDS
2. Updating the app and creating a new controller/view so that it can be deployed via cloud.gov


jduss4 (approved):

jduss4:
Tests are not passing locally but I'm okay with merging this for now and opening a new issue to investigate how to get tests to work with the asset pipeline. I'm not sure why they would work with cloud.gov but since they are I think an issue is okay so we can get cloud.gov running off the main branch.

jduss4:
Well, okay workaround for now is `bundle exec rake assets:precompile RAILS_ENV=test` but ideally we shouldn't need to compile them at all

jduss4 (approved):

Sgtpluck:
I think the tests were failing because of timeout issue -- it takes awhile for uswds to compile. Once it's been compiled, it seems like the test run fine. We should definitely investigate the issue and see if there's a better way to get request specs to run!

PR #15, opened 2021-07-23
Add test workflow

Sgtpluck:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,67 @@
...
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - uses: actions/checkout@v2
+      
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+      
+      - name: Install dependencies
```
Sgtpluck, line 41: should we figure out how to cache these and the precompiled assets?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,67 @@
...
+          POSTGRES_DB: complaint_tracker_test
+          POSTGRES_USER: complaint_tracker 
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_PORT: ${{ job.services.postgres.ports[5432] }}
+        run: |
+          cp config/database.ci.yml config/database.yml
+          bundle exec rake db:create db:migrate
+
+      - name: Precompile assets
+        run: bundle exec rake assets:precompile
```
rahearn, line 57: I think we can skip this step. It's not needed (normally) for tests, and the circleci ruby buildpack will do it on deploy.

.github/workflows/tests.yml
rahearn: Yes, but I don't think it needs to be done as part of this PR.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,67 @@
...
+      
+      - name: Run tests
+        env:
+          RAILS_ENV: test
+          POSTGRES_HOST: localhost
+          POSTGRES_DB: complaint_tracker_test
+          POSTGRES_USER: complaint_tracker 
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_PORT: ${{ job.services.postgres.ports[5432] }}
```
rahearn, line 66: Can we build and pass a `DATABASE_URL` env variable here (and in the create/migrate step too)? `DATABASE_URL` always overrides whatever is in `database.yml`, so we wouldn't need to have a separate ci file that needs to be copied over.

.github/workflows/tests.yml
Sgtpluck: the request specs seem to require the uswds assets, unfortunately. they're new to me, so i definitely might be setting them up wrong, but if we don't precompile, then the tests choke on 
`@import 'uswds/dist/scss/uswds'`

.github/workflows/tests.yml
Sgtpluck: sure. is there any documentation you can point me to re: DATABASE_URL? i didn't come across it when i was trying to figure this out!

rahearn: https://edgeguides.rubyonrails.org/configuring.html#configuring-a-database section 3.19 and 3.20 go into it.

.github/workflows/tests.yml
rahearn: I haven't used request specs much before, so disregard my comment!

.github/workflows/tests.yml
Sgtpluck: thanks!!

rahearn (approved):
🎉

Issue #14, opened 2021-07-23
githook


Sgtpluck:
User Story:
As a dev writing code
I want a githook to run the formatter and linter when I add a commit
So that I don't have to worry about my formatting!

Acceptance:
Given I am a developer writing code
When I commit the code
The linter and formatter should run.

PR #16, opened 2021-07-23
adds basic git hook on pre-commit to run linter

jduss4:
runs the linter
adds documentation for git hook setup and bypass

closes https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/14

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,22 @@
+#! /usr/bin/env bash
+#
+# This hook runs on `git commit` and will prevent you from committing without
+# approval from the linter and tests.
+#
+# To run, this file must be symlinked to:
+# .git/hooks/pre-commit
+#
+# To bypass this hook, run:
+# $ git commit --no-verify
+# $ git commit -n
+
+echo "Running linter..."
+bundle exec rake standard
+linter_status=$?
+
+if [ "${linter_status}" = 0 ]; then
+    exit 0
+else
+    echo "Fix above before committing. Run 'git commit -n' to bypass linter."
+    exit 1
+fi
--- a/README.md
+++ b/README.md
@@ -20,12 +20,30 @@ visibility into status and actions taken to address a complaint.
 * Run migrations: `bundle exec rake db:migrate`
 * Run the server: `bundle exec rails s`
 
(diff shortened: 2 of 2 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1,26 @@
...
+linter_status=$?
+
+echo "Running tests..."
+bundle exec rake spec
+test_status=$?
+
+if [ "${linter_status}" = 0 ] && [ "${test_status}" = 0 ]; then
+    exit 0
+else
+    echo "Roll your lint and fix your tests before committing!"
```
Sgtpluck, line 24: [question] Hmm, running the tests before every commit will probably get annoying when we have a lot of tests. would it make more sense to run the linters and just let CI deal with tests? (i guess since it's a symlinked file everyone can decide for themselves)

jduss4: That's true! That works for me if you all think it's a good idea.

Sgtpluck: i think that's a good idea. we could have a pre-push githook if you think it's important to run tests locally before pushing! (although realistically, this github action that i am trying to set up should be guard enough for tests.)

.githooks/pre-commit
```diff
@@ -0,0 +1,26 @@
+#! /usr/bin/env sh
```
jduss4, line 1: maybe I should make this `bash`, actually

README.md
```diff
@@ -20,13 +20,31 @@ visibility into status and actions taken to address a complaint.
...
 ### Running Tests
 
 Ruby linter: `bundle exec rake standard`
 Tests: `bundle exec rake spec`
 
 Run everything: `bundle exec rake`
 
+### Set up automatic linting and testing
+
+If you would like the linter and tests to run before you commit, symlink `.githooks/pre-commit` to `.git/hooks/pre-commit`.
```
Sgtpluck, line 34: [tiny nit] just linter

Sgtpluck (approved):
just one tiny thing in the README! feel free to merge once that's updated.

README.md
jduss4: good catch!

PR #19, opened 2021-07-28
Add pa11y workflow

Sgtpluck:
I thought it would be good to have automated accessibility testing set up early.

This workflow does not currently post the pa11y output into a GitHub comment, although I could set that up. Would that be helpful?


.pa11yci
```diff
@@ -0,0 +1,10 @@
+{
+  "defaults": {
+    "concurrency": 4,
+    "standard": "WCAG2AA",
+    "runners": ["axe"]
+  },
+  "urls": [
+    "http://localhost:3000"
```
Sgtpluck, line 8: we may need to futz with this as we add more on

rahearn:
> This workflow does not currently post the pa11y output into a GitHub comment, although I could set that up. Would that be helpful?

That sounds helpful, would it be a comment on the PR that it is testing?

rahearn (approved):
🎉 I agree on the early a11y testing. Thank you for setting it up!

jduss4 (approved):

PR #22, opened 2021-07-28
Limit to chokidar version that fixes globparent vulnerability

Sgtpluck:
Will this break something??


rahearn (approved):
I can't figure out how to break it

Issue #9, opened 2021-07-22
Wire in authentication


Sgtpluck:
As a PS,
I want to be able to log in with my HSES credentials 
so that I don't have to remember any other passwords

Acceptance Criteria:
* Navigate to $STAGING-URL
* There should be screen that directs you to log in via HSES with a button
* Hit the button
* Log in with HSES credentials
You should be redirected to a landing page that will eventually house the summary view of PS's complaints.

PR #38, opened 2021-07-30
Add authentication with HSES

rahearn:
Paired with @jduss4 on simple login/logout.

Original plan was to store the full auth response in the session, but my user response was large enough to overflow the max cookie size. We'll need to store that info either in postgres or cache it in redis if we decide it is useful to keep around.

login/logout button styling is... not present.

#9


spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+
+RSpec.describe ApplicationHelper, type: :helper do
+  describe "#current_user" do
+    it "returns the username" do
```
Sgtpluck, line 5: this test doesn't feel like it's testing much. i think it would be a more solid test if we set `session["username"]` and then called `helper.current_user`

```
describe "#current_user" do
  it "returns the username" do
    session["username"] = "Test User"
    expect(helper.current_user).to eq "Test User"
  end
end
```

doc/adr/0003-authentication.md
```diff
@@ -0,0 +1,27 @@
+# 3. Authentication
```
Sgtpluck, line 1: oooh yay

Sgtpluck (approved):
I have one suggestion about the tests, but otherwise this looks great!

One other question I had was that I was under the understanding that there was a login page that came along with this auth (that looks like the TTAHub login page) -- looks like that's not the case. We should probably add a card to route an unauthed user to a login view, yes?

rahearn:
> We should probably add a card to route an unauthed user to a login view, yes?

Yes, or a card for a new unauthenticated landing page at the `root_path`

spec/helpers/application_helper_spec.rb
rahearn: Good idea

doc/adr/0003-authentication.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Users of the complaint tracker need to login for access to sensitive data.
+
+All users will already have an HSES account, and HSES is the main sore of complaint data currently.
```
jduss4, line 13:
```suggestion
All users will already have an HSES account, and HSES is the main source of complaint data currently.
```

PR #41, opened 2021-08-02
Add username

Sgtpluck:


rahearn (changes requested):
Code looks good, but the tests need to be updated. Also, is this completing a particular story? I'd like to see the link in the PR description

spec/helpers/application_helper_spec.rb
```diff
@@ -2,9 +2,14 @@
 
 RSpec.describe ApplicationHelper, type: :helper do
   describe "#current_user" do
+    session_user = {
+      name: "Test User",
+      email: "testuser@example.com"
+    }
```
rahearn, line 8: I'd generally use a `let` block here in rspec. Is there an advantage to just creating a bare variable in this scope?

spec/requests/complaints_spec.rb
```diff
@@ -7,9 +7,26 @@
...
+      end
+    end
+
+    describe "with an authorized user" do
+      before do
+        user = {
+          name: "Test User",
+          email: "testuser@example.com"
+        }.with_indifferent_access
+        allow_any_instance_of(ActionDispatch::Request::Session).to receive(:[]).and_return user
```
rahearn, line 23:
```suggestion
        allow_any_instance_of(ActionDispatch::Request::Session).to receive(:[]).with("user").and_return user
```

that feels like it might be more future-proof if/when we add other things to the session.

spec/helpers/application_helper_spec.rb
Sgtpluck: i don't think so, i'll update it to a `let`!

Sgtpluck:
@rahearn ready for one more review when you have time!

app/controllers/sessions_controller.rb
```diff
@@ -1,6 +1,9 @@
 class SessionsController < ApplicationController
   def create
-    session["username"] = auth_info["uid"]
+    session["user"] = {
+      email: auth_info["uid"],
```
jduss4, line 4: Suggestion:  I'm not sure this is the email, but rather is the username?  Would you mind switching this back or adding username and email? This is what the debug line below outputs:

```
"uid": "test.rosps",
  "info": {
    "name": "Test ID RO Supervisory Program Specialist",
    "email": "33212@hsesinfo.org",

```

rahearn: Relatedly, the username field in HSES is often, but not always, actually an email address.

Sgtpluck: I'd be happy to change the field to `uid`; I'd rather not have a name and a username because that's confusing to me. (I'm also curious if, in reality, the uid is ever not the user's email address?)

On reflection, I don't think we currently need their email anyways; if we ever get into sending notifications we will, but currently a ways away from that.

rahearn: > (I'm also curious if, in reality, the uid is ever not the user's email address?)

In production, yes, the uid is often not an email address. See: https://github.com/HHS/Head-Start-TTADP/blob/ce6f7d5a22e51949de680c31a644c9bb9dcb01b6/src/app.js#L80-L83 for how TTA Hub assigns the user email address to their saved HSES users some of the time.

rahearn (approved):
🎉

app/controllers/sessions_controller.rb
Sgtpluck: cool, thanks for those details!

Issue #21, opened 2021-07-28
Prepare Open API documentation to request API endpoint from HSES


jduss4:
User Story:
As an engineer
I want to create an Open API yaml file with the fields we wish to request from HSES
So that they will have clarity around the API requests / responses we would like to use to access their data

Acceptance:
- [x] we have an Open API file which validates and
- [x] has all the fields we want in it
- [x] has the basic API behavior we may need (pagination, filtering)

jduss4:
Not sure about the sensitivity of the fields, so putting this work in a [google doc](https://docs.google.com/document/d/1j0ICytviL0QXM5sYBCeAH0-6Nhj0Y5OyiSz6H7VZHx0/edit#) for now.

PR #52, opened 2021-08-10
Complaints fake data

### LINKED ISSUE #49: Fake /issues endpoint
User Story:
As a developer
I want to create fake data imitating our ideal HSES response for `/issues` and make it available to the app
So that we can provide information to be used in the user interface

Acceptance:
- [ ] fake JSON data exists which matches parameters of #21 
- [ ] can access data from a controller

jduss4:
## Description of change

Adds mocked in data imitating future HSES response
Addresses: https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/49

## Acceptance Criteria

- [ ] fake JSON data is created which matches parameters of #21
- [ ] data can be accessed from a controller

## How to test

`rails spec` runs a couple cursory tests which check the basic structure of the data being created

If you want to look at the data being generated, you can add `<%= @complaints.to_json %>` to `app/views/complaints/index.html.erb`

## Issue(s)

- closes #49

## Checklist

- [X] Meets issue criteria
  - moved to "in progress"
  - for reviewers: do you think this is enough testing?
- [X] Meets accessibility standards (WCAG 2.1 Levels A, AA)
  - for reviewers: should I make an ADR explaining why we went with ffaker instead of the staging export data?

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/lib/fake_data/complaints.rb
+++ b/app/lib/fake_data/complaints.rb
@@ -1,0 +1,73 @@
+require "ffaker"
+
+module FakeData::Complaints
+  def self.data
+    {
+      id: FFaker::Random.rand(9999).to_s,
+      type: "issues",
+      attributes: {
+        issueLastUpdated: FFaker::Time.date,
+        creationDate: FFaker::Time.date,
+        closedDate: FFaker::Time.date,
+        reopenedDate: FFaker::Time.date,
+        initialContactDate: FFaker::Time.date,
+        type: FFaker::Random.rand(2),
+        otherType: FFaker::Lorem.word,
+        issue: FFaker::Lorem.phrase,
+        priority: FFaker::Random.rand(3),
+        status: FFaker::Random.rand(3),
+        dueDate: FFaker::Time.date,
+        # these just sound cool for use as the grantee name :)
+        grantee: FFaker::AddressUK.street_name
+      },
+      relationships: {
+        grantAward: {
+          meta: {
+            id: FFaker::Random.rand(9999).to_s
+          }
+        },
+        grantProgram: {
+          meta: {
+            id: FFaker::Random.rand(999).to_s
+          }
+        },
+        region: {
+          meta: {
+            id: FFaker::Random.rand(10).to_s
+          }
+        }
+      },
+      links: {
+        self: "https://example.com/TODO",
+        html: "https://example.com/TODO"
+      }
+    }
+  end
+
+  def self.generate_response
+    json = jsonapi_wrapper
+    25.times do |item|
+      json[:data] << data
+    end
+    json
+  end
+
+  def self.jsonapi_wrapper
+    {
+      meta: {
+        pageNumber: 1,
+        pageSize: 25,
+        pageTotalCount: 1,
+        itemTotalCount: 25
+      },
+      data: [],
+      links: {
+        self: "https://example.com/TODO",
+        first: "https://example.com/TODO",
+        last: "https://example.com/TODO",
+        prev: "https://example.com/TODO",
+        next: "https://example.com/TODO"
+      }
+    }
+  end
+end
--- a/app/lib/fake_data/complaints.rb
+++ b/app/lib/fake_data/complaints.rb
@@ -3,14 +3,14 @@ require "ffaker"
 module FakeData::Complaints
   def self.data
     {
-      id: FFaker::Random.rand(9999).to_s,
+      id: identifier,
       type: "issues",
       attributes: {
-        issueLastUpdated: FFaker::Time.date,
-        creationDate: FFaker::Time.date,
-        closedDate: FFaker::Time.date,
-        reopenedDate: FFaker::Time.date,
-        initialContactDate: FFaker::Time.date,
+        issueLastUpdated: datetime,
+        creationDate: datetime,
+        closedDate: datetime,
+        reopenedDate: datetime,
+        initialContactDate: datetime,
         type: FFaker::Random.rand(2),
         otherType: FFaker::Lorem.word,
         issue: FFaker::Lorem.phrase,
@@ -23,17 +23,17 @@ module FakeData::Complaints
       relationships: {
         grantAward: {
           meta: {
-            id: FFaker::Random.rand(9999).to_s
+            id: identifier
           }
         },
         grantProgram: {
           meta: {
-            id: FFaker::Random.rand(999).to_s
+            id: identifier
           }
         },
         region: {
           meta: {
-            id: FFaker::Random.rand(10).to_s
+            id: identifier
           }
         }
       },
@@ -42,6 +42,17 @@ module FakeData::Complaints
         html: "https://example.com/TODO"
       }
     }
+  end
+
+  def self.datetime
+    faketime = FFaker::Time.datetime
+    # format UTC date YYYY-MM-DDTHH:MM:SSZ
+    faketime.strftime("%FT%TZ")
+  end
+
+  # returns a string because actual identifiers may not be integers
+  def self.identifier
+    FFaker::Random.rand(9999).to_s
   end
 
   def self.generate_response
--- a/spec/lib/fake_data/complaints_spec.rb
+++ b/spec/lib/fake_data/complaints_spec.rb
@@ -6,6 +6,7 @@ RSpec.describe FakeData::Complaints do
       complaint = FakeData::Complaints.data
 
       expect(complaint[:id]).to_not be_nil
+      expect(complaint[:id]).to be_a(String)
       expect(complaint[:attributes]).to be_a(Hash)
       expect(complaint.dig(:relationships, :grantAward, :meta, :id)).to be_a(String)
       expect(complaint[:links]).to be_a(Hash)
@@ -15,20 +16,17 @@ RSpec.describe FakeData::Complaints do
   describe "datetime" do
     it "returns a string formatted as ISO-8601 UTC datetime" do
       dt = FakeData::Complaints.datetime
+
       expect(dt).to be_a(String)
-      begin
-        expect(Time.iso8601(dt)).to be_a(Time)
-      rescue
-        expect(false).to be_truthy
-      end
+      expect(Time.iso8601(dt)).to be_a(Time)
     end
   end
 
   describe "identifier" do
-    it "returns a string of numbers" do
+    it "returns a string" do
       id = FakeData::Complaints.identifier
+
       expect(id).to be_a(String)
-      expect(id.to_i.to_s).to eq(id)
     end
   end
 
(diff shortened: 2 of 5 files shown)


spec/lib/fake_data/complaints_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+
+RSpec.describe FakeData::Complaints do
+  describe "jsonapi_wrapper" do
+    it "should return a object with JSON-API top level JSON API" do
```
Sgtpluck, line 5: [extreme nitpick] testing philosophies have moved away from "should" language. 

`it returns an object that matches JSON API specifications`

app/lib/fake_data/complaints.rb
```diff
@@ -0,0 +1,72 @@
+require "ffaker"
+
+module FakeData::Complaints
+  def self.complaint_data
```
Sgtpluck, line 4: [suggestion] can we rename this to `data`?

app/lib/fake_data/complaints.rb
```diff
@@ -0,0 +1,72 @@
...
+      },
+      links: {
+        self: "TODO",
+        html: "TODO"
+      }
+    }
+  end
+
+  def self.generate_response
+    json = jsonapi_wrapper
```
Sgtpluck, line 47: [thought/nonblocking] if we were leaving this in for longer, i would suggest making a class because the fact that we're creating a wrapper object here and then applying changes to it and then returning it suggests the need to me. (also a bunch of related `self` methods is always a bit of a smell to me.) 

since it's a short-lived one-off, it's fine how it is, but i wanted to raise it just as something to think about for the future!

spec/lib/fake_data/complaints_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      wrapper = FakeData::Complaints.jsonapi_wrapper
+
+      expect(wrapper[:meta]).to be_a(Hash)
+      expect(wrapper[:data]).to be_a(Array)
+      expect(wrapper[:links]).to be_a(Hash)
+    end
+  end
+
+  describe "complaint_data" do
+    it "should return a single complaint object" do
```
Sgtpluck, line 15: `returns a single complaint object`

spec/lib/fake_data/complaints_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      complaint = FakeData::Complaints.complaint_data
+
+      expect(complaint[:id]).to_not be_nil
```
Sgtpluck, line 18: [suggestion] do we want to expect that this is an integer?

Sgtpluck (changes requested):
A couple suggestions, and some thoughts for the 💫 future 💫

spec/lib/fake_data/complaints_spec.rb
Sgtpluck: cool, let's expect it to be the thing it is!

app/lib/fake_data/complaints.rb
```diff
@@ -0,0 +1,72 @@
+require "ffaker"
+
+module FakeData::Complaints
+  def self.complaint_data
+    {
+      id: FFaker::Random.rand(9999),
+      attributes: {
+        issueLastUpdated: FFaker::Time.date,
```
rahearn, line 8: [question] in the response, the dates [should be ISO 8601 formatted](https://jsonapi.org/recommendations/#date-and-time-fields). Do we want to ensure that format here? Or rely on the later `to_json` call to do it for us?

app/lib/fake_data/complaints.rb
```diff
@@ -0,0 +1,72 @@
...
+        closedDate: FFaker::Time.date,
+        reopenedDate: FFaker::Time.date,
+        initialContactDate: FFaker::Time.date,
+        type: FFaker::Random.rand(2),
+        otherType: FFaker::Lorem.word,
+        issue: FFaker::Lorem.phrase,
+        priority: FFaker::Random.rand(3),
+        status: FFaker::Random.rand(3),
+        dueDate: FFaker::Time.date,
+        # these just sound cool for use as the grantee name :)
```
rahearn, line 19: 😆 ❤️

app/lib/fake_data/complaints.rb
```diff
@@ -0,0 +1,72 @@
...
+        grantee: FFaker::AddressUK.street_name
+      },
+      relationships: {
+        grantAward: {
+          meta: {
+            id: FFaker::Random.rand(9999)
```
rahearn, line 25: [suggestion] This is less of a `must` than the top level `id` since it's in `meta` and non-standard, but I think we should use strings here and in the other relationships for the same reasons.

app/lib/fake_data/complaints.rb
```diff
@@ -0,0 +1,72 @@
...
+          }
+        },
+        region: {
+          meta: {
+            id: FFaker::Random.rand(10)
+          }
+        }
+      },
+      links: {
+        self: "TODO",
```
rahearn, line 40: [thought] should these be valid URLs? something like `https://example.com/TODO` just in case someone at some point tries parsing it?

config/routes.rb
```diff
@@ -3,5 +3,5 @@
   get "/", to: "complaints#index", as: :root
 
   get "/oauth2-client/login/oauth2/code/", to: "sessions#create"
-  get "logout", to: "sessions#destroy"
+  get "/logout", to: "sessions#destroy"
```
rahearn, line 6: [praise] good catch!

rahearn (changes requested):
Can you please merge in main to get the latest github actions checks.

spec/lib/fake_data/complaints_spec.rb
jduss4: Okay!  I was using the type from HSES's issues table, but I think it makes sense to be a string for flexibility.

app/lib/fake_data/complaints.rb
jduss4: Yeah, that's fair.  I had started making a class at first when I wanted to pass in pagination / filters and then decided against doing too much for a little fake data.  We will probably want to rethink the structure anyway if we want to mock a single complaint in the near future, since we would want to reuse the `data` function.

spec/lib/fake_data/complaints_spec.rb
jduss4: I will update the Open API spec for the identifiers, also

app/lib/fake_data/complaints.rb
```diff
@@ -0,0 +1,72 @@
+require "ffaker"
+
+module FakeData::Complaints
+  def self.complaint_data
+    {
+      id: FFaker::Random.rand(9999),
```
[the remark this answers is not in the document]
jduss4: Oh no, how did I miss type?!  Thanks

app/lib/fake_data/complaints.rb
jduss4: I may be misunderstanding your question, but [FFAker's Time](https://github.com/ffaker/ffaker/blob/main/REFERENCE.md#ffakertime) `date` returns `YYYY-MM-DD`, which is ISO 8601.  I do like the idea of specifying in the Open API description for each date that we would like `YYYY-MM-DD` returned, though, to ensure that's what we get back from HSES.

app/lib/fake_data/complaints.rb
jduss4: Yeah, I think if we're making one id a string, they all ought to be strings

app/lib/fake_data/complaints.rb
jduss4: Works for me!

app/lib/fake_data/complaints.rb
rahearn: Does FFaker return a string that's `YYYY-MM-DD` or a ruby `Date` object? I assumed the latter.

Also, if the time is available this would likely be better as a datetime (more data is better than less) in which case 8601 is: `YYYY-MM-DDTHH:MM:SSZ` (assuming that the time is in UTC timezone, and also I forget how many subseconds are returned)

jduss4: Oh nooo you are correct, I was fooled by their documentation just listing sample dates as though they were strings.  I will mess with this, thanks for catching it.

spec/lib/fake_data/complaints_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        expect(false).to be_truthy
+      end
+    end
+  end
+
+  describe "identifier" do
+    it "returns a string of numbers" do
+      id = FakeData::Complaints.identifier
+      expect(id).to be_a(String)
+      expect(id.to_i.to_s).to eq(id)
```
Sgtpluck, line 31: i'd prefer this to be 

`expect(id.to_i).not_to eq 0` (assuming this HAS to be a string of numbers)

Sgtpluck (changes requested):

spec/lib/fake_data/complaints_spec.rb
jduss4: It doesn't have to be but it currently is just to populate it with something

spec/lib/fake_data/complaints_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  describe "datetime" do
+    it "returns a string formatted as ISO-8601 UTC datetime" do
+      dt = FakeData::Complaints.datetime
+      expect(dt).to be_a(String)
+      begin
+        expect(Time.iso8601(dt)).to be_a(Time)
+      rescue
```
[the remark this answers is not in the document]
jduss4: Turns out I didn't know how rspec works, will fix

spec/lib/fake_data/complaints_spec.rb
jduss4: I don't think that `expect(id.to_i).not_to eq 0` works entirely, because `"12".to_i` returns 12, but also it doesn't really matter what the value of this is.  Maybe I'll just remove the test checking the substance and be happy that there is an identifier which is a string

spec/lib/fake_data/complaints_spec.rb
rahearn: This test is still missing the verification that `[:id]` is a string

spec/lib/fake_data/complaints_spec.rb
rahearn: +1 to not worrying about the substance of this identifier.

rahearn (approved):
🎉

Sgtpluck (approved):

PR #53, opened 2021-08-11
Update header to reflect design

### LINKED ISSUE #50: Update header
As a Program Specialist I want to see my name and the name of the Complaints Tracker in the header
So that I know that I am logged into the Complaints Tracker

Acceptance:
Given that I am a Program Specialist
When I login to Complaints Tracker
Then I can see My name and "Complaints Tracker" in the Header

Design:
https://www.figma.com/file/k8sPnWItmiBAyF4cw6Uuog/OHS-Complaints-Tracker?node-id=0%3A1

Sgtpluck:
## Description of change
This change reflects the design in https://www.figma.com/file/k8sPnWItmiBAyF4cw6Uuog/OHS-Complaints-Tracker?node-id=43%3A19740

Screen Shot 2021-08-11 at 8 33 52 AM

Screen Shot 2021-08-11 at 8 34 10 AM

Notes: Roles are not added, the name of the staging user is "Test ID RO Supervisory Program Specialist"

## Acceptance Criteria
Given that I am a Program Specialist
When I login to Complaints Tracker
Then I can see My name and "Complaints Tracker" in the Header

## Issue(s)
* closes #50

## Checklist

- [x] Meets issue criteria
- [ ] Project board status updated
- [NA] API Documentation updated
- [NA] Boundary diagram updated
- [NA] Logical Data Model updated
- [NA] [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


app/views/application/_header.html.erb
```diff
@@ -1,18 +1,25 @@
 <header class="usa-header usa-header--basic">
-  <div class="usa-nav-container">
-    <div class="usa-navbar">
```
Sgtpluck, line 3: the actual USWDS header component is very focused around navigation, so I moved towards using the grid and flexbox classes to make this look right. There may/probably are better ways to do this? I don't know!

rahearn: I think this is a fine way to do it as long as we remember to switch back to a navbar if/when we add some navigation to the header.

app/views/application/_header.html.erb
```diff
@@ -1,18 +1,25 @@
...
             OHS Complaint Tracker
-          </a>
         </em>
       </div>
-      <% if user_signed_in? %>
-        <%= link_to "Logout #{current_user["uid"]}", logout_path, class: "usa-button", data: {turbolinks: false} %>
-      <% else %>
-        <%= button_to "Login", "/auth/hses", class: "usa-button" %>
-      <% end %>
+      <div class="grid-row grid-col-6 grid-gap flex-align-center flex-justify-end">
```
rahearn, line 9: [question] I thought that grid and flexbox were two approaches to solving layout problems, and that they weren't really meant to be mixed together. Are there `grid-align-center grid-justify-end` classes? I know those types of things exist in tailwind css but I'm not sure about uswds.

spec/views/application/_header.html.erb_spec.rb
```diff
@@ -0,0 +1,26 @@
+require "rails_helper"
+
+RSpec.describe "rendering header buttons do" do
+  describe "with no authorized user logged in" do
+    it "has a logout button" do
+      render partial: "application/header.html.erb"
```
rahearn, line 6: [praise] I love this approach of testing just the partial to verify this behavior. I always get my view tests way too bogged down in setup by testing the whole page.

app/views/application/_header.html.erb
```diff
@@ -1,18 +1,25 @@
...
+        <% if user_signed_in? %>
+          <div class="grid-col-auto">
+            <%= current_user["name"] %>
+          </div>
+          <div class="grid-col-auto">
+            <%= button_to "Log out", logout_path, class: "usa-button usa-button--accent-warm", data: {turbolinks: false}, method: :get %>
```
rahearn, line 15: [thought] since we're using `button_to` anyway, should we switch the logout route to be a `delete` method call, or keep it as `get`?

`delete` feels more pure `REST` but I always wonder what the value of that actually is for non API routes.

rahearn (approved):
👍🏻 

The one thing to really look at again is the flex css classes in the header. Otherwise (and even with that) this is good.

app/views/application/_header.html.erb
Sgtpluck: they are USWDS classes.
https://designsystem.digital.gov/utilities/flex/

The examples all use `grid-row` and `grid-col` classes along with the flex classes.

I did do some experimenting to try to remove the `grid-col-6` class but in terms of the design layout, it seemed much cleaner to group the different elements by the column-width and then flex within that.

app/views/application/_header.html.erb
Sgtpluck: that's fine by me! i assumed there had been a Reason(tm) you had made it a `get`!

Sgtpluck: i changed it to a button for accessibility reasons -- it looks and acts like a button so it should be a button!

app/views/application/_header.html.erb
rahearn: Huh, yeah I don't see anything about alignment or justify in the grid docs, so I guess this is the preferred way to do it in USWDS.

Issue #79, opened 2021-08-19
Persistent Fake Data


jduss4:
User Story:
As a developer / tester of this software
I want to have semi-persistent fake data
So that the information will not change as I use the site

Acceptance:
Given that I am a developer / tester
When I refresh the index page,
The data displayed will remain the same

PR #80, opened 2021-08-19
Adds class variable to controller to persist fake issues data

jduss4:
## Description of change

Adds a class variable to the ComplaintsController which contains the fake data. This way, the fake data will not change when the user refreshes the page or navigates between (future) complaints pages.

## Acceptance Criteria

The fake data does not change on every request or between users, just on app reload.

## How to test

Manually, can start up the app and refresh the index page a few times to verify that the fake data is not changing. In terms of how to test the data is persistent in rspec, I am seeking advice.

## Issue(s)

* closes https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/79

## Checklist

- [X] Meets issue criteria
- ~Meets accessibility standards (WCAG 2.1 Levels A, AA)~
- ~API Documentation updated~
- ~Boundary diagram updated~
- ~Logical Data Model updated~
- ~[Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions~

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/controllers/complaints_controller.rb
+++ b/app/controllers/complaints_controller.rb
@@ -1,7 +1,5 @@
 class ComplaintsController < ApplicationController
-  @@hses_issues = FakeData::ApiResponse.generate_hses_issues_response[:data]
-
   def index
-    @complaints = @@hses_issues
+    @complaints = FakeData::ApiResponse.hses_issues_response[:data]
   end
 end
--- a/app/controllers/complaints_controller.rb
+++ b/app/controllers/complaints_controller.rb
@@ -1,5 +1,7 @@
 class ComplaintsController < ApplicationController
+  @@hses_issues = FakeData::ApiResponse.generate_hses_issues_response[:data]
+
   def index
-    @complaints = FakeData::ApiResponse.generate_hses_issues_response[:data]
+    @complaints = @@hses_issues
   end
 end
--- a/spec/models/api_response_spec.rb
+++ b/spec/models/api_response_spec.rb
@@ -1,11 +1,20 @@
 require "rails_helper"
 
 RSpec.describe FakeData::ApiResponse do
-  describe "generate_response" do
+  describe "generate_hses_issues_response" do
     it "returns a JSON-API object with 25 complaints" do
       res = FakeData::ApiResponse.generate_hses_issues_response
 
       expect(res[:data].length).to eq(25)
+    end
+  end
+
+  describe "hses_issues_response" do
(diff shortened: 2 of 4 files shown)


app/controllers/complaints_controller.rb
```diff
@@ -1,5 +1,7 @@
 class ComplaintsController < ApplicationController
+  @@hses_issues = FakeData::ApiResponse.generate_hses_issues_response[:data]
```
rahearn, line 2: [suggestion] I think I like the method of saving these, but I think I don't like the controller being responsible for holding onto the data. What if we added a `FakeData::ApiResponse.hses_issues_response` method that memoized the return value of `generate_hses_issues_response`?

I think that would also clarify where in the test stack is the right place to test the behavior.

spec/models/api_response_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe FakeData::ApiResponse do
   describe "generate_response" do
     it "returns a JSON-API object with 25 complaints" do
-      res = FakeData::ApiResponse.generate_hses_issues_response
+      res = FakeData::ApiResponse.hses_issues_response
```
rahearn, line 6: [suggestion] I think I'd keep this test on the generate method, and then add another one that tests the memoization.

rahearn (commented):
I believe this should work well, but have you tried testing through just reloading the browser?

jduss4:
> I believe this should work well, but have you tried testing through just reloading the browser?

Yes sir!

app/controllers/complaints_controller.rb
jduss4: Ope yep, that is a very good call

spec/models/api_response_spec.rb
jduss4: I responded to the wrong comment, but yes that is a good idea!

Sgtpluck: [suggestion] i think you can remove lines 12-17, since you kept it on the generate method

Sgtpluck (approved):
just a little deletion request!

spec/models/api_response_spec.rb
jduss4: It's true, I decided to put it in to make sure that the method was doing what it said it would do (return issues) but it's definitely a duplicate test since there's not in terms of logic that would differ.  I can take it out!

PR #86, opened 2021-08-23
Update issue templates

Sgtpluck:
Adding a task template

### DIFF
--- a/.github/ISSUE_TEMPLATE/task.md
+++ b/.github/ISSUE_TEMPLATE/task.md
@@ -1,0 +1,20 @@
+---
+name: Task
+about: When a user story has to be broken into smaller engineering chunks, use a task!
+title: 'Task: Part of #Issue'
+labels: ''
+assignees: Sgtpluck
+
+---
+
+This task is part of #71
+
+Full User Story:
+As a Program Specialist
+I want my new or urgent complaints to have a color coding
+So that I know what complaints have an urgent need to be addressed
+
+This task is scoped to:
+- [] 
+- [] 
+- []
--- a/.github/ISSUE_TEMPLATE/task.md
+++ b/.github/ISSUE_TEMPLATE/task.md
@@ -2,17 +2,17 @@
 name: Task
 about: When a user story has to be broken into smaller engineering chunks, use a task!
 title: 'Task: Part of #Issue'
-labels: ''
-assignees: Sgtpluck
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/task.md
```diff
@@ -0,0 +1,20 @@
...
+title: 'Task: Part of #Issue'
+labels: ''
+assignees: Sgtpluck
+
+---
+
+This task is part of #71
+
+Full User Story:
+As a Program Specialist
```
jduss4, line 13: [suggestion] It seems fine to use a realistic example, but if we wanted to match the other templates you could just copy the user-story example entirely here (`Given that I am a [type of user]`)

.github/ISSUE_TEMPLATE/task.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Task
+about: When a user story has to be broken into smaller engineering chunks, use a task!
+title: 'Task: Part of #Issue'
+labels: ''
+assignees: Sgtpluck
```
jduss4, line 6: [question] What are the odds that you end up getting accidentally assigned a bunch of tasks because you put your name in the template?  :)

.github/ISSUE_TEMPLATE/task.md
```diff
@@ -0,0 +1,20 @@
...
+
+This task is part of #71
+
+Full User Story:
+As a Program Specialist
+I want my new or urgent complaints to have a color coding
+So that I know what complaints have an urgent need to be addressed
+
+This task is scoped to:
+- []
```
jduss4, line 18: [suggestion] if you are adding a real user story as an example, you could consider adding an example task below, too, if you want to

jduss4 (approved):
A few thoughts / suggestions but nothing holding up the PR!

.github/ISSUE_TEMPLATE/task.md
Sgtpluck: ugh i removed myself from it, i think there is a little bug in their creation system, i'll manually take myself out of this

.github/ISSUE_TEMPLATE/task.md
Sgtpluck: yeah that's a good idea

jduss4 (approved):

PR #88, opened 2021-08-24
Refactor of Api to make it flexible for future

Sgtpluck:
## Description of change
This change is a proposal! 

The current structure we have for the Api response is not very OO, and will not scale very well. I have done a preliminary refactor that I believe will be more flexible going forward.

I propose having FakeData and real Api classes that mirror each other's structure so that we can do a basic switch on environments to access the real data (when that is available) in staging and production and the fake data when we are testing or in development. This will also give us a lot of flexibility in the real API to switch from staging to production endpoints and credentials.

I also switched the FakeData caching to a Singleton -- that is a little cleaner than a global variable as it has to be required into a file, and also will allow data to persist between app restarts in a little more realistic way.

If this proposal meets everyone's approval, I will create an ADR to describe why we took this approach.

## How to test
Click around on the site and make sure the index view and detail view still work.

## Checklist

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


app/models/api/fake_data/hses.rb
```diff
@@ -0,0 +1,25 @@
+require "fake_api_response_wrapper"
+require "fake_issues"
+class Api::FakeData::Hses
+  class Issue
+    include FakeApiResponseWrapper
+    attr_accessor :id
+
+    def initialize(options_arr)
+      options = Hash[*options_arr]
```
Sgtpluck, line 9: i don't love this, but couldn't figure out a better way to optionally pass in arguments (the double splat operator would require creating an initializer for Issues, which also didn't feel great.

very open to different ways of handling the optional arguments being passed into our Api wrapper!

app/models/api/hses.rb
```diff
@@ -0,0 +1,26 @@
+require "fake_api_response_wrapper"
+require "fake_issues"
+
+class Api::Hses
```
Sgtpluck, line 4: this is here to ensure that staging still works! once we have a working endpoint, we can transition this to our real HTTP request code.

lib/fake_api_response_wrapper.rb
```diff
@@ -0,0 +1,28 @@
+module FakeApiResponseWrapper
```
Sgtpluck, line 1: this is now a module that could be mixed in with different fake system Api wrappers, since we anticipate this being consistent.

app/models/api.rb
```diff
@@ -0,0 +1,11 @@
+class Api
+  def self.request(system, endpoint, args = nil)
+    "Api::#{Api.env}#{system.capitalize}::#{endpoint.capitalize}".constantize.new(*args).request
+  end
+
+  def self.env
+    if Rails.env == "development" || Rails.env == "test"
+      "FakeData::"
```
jduss4, line 8: This is a clever way of doing this, although I don't love a giant string + `constantize`.  Unfortunately, I can't think of a better way of doing it beyond like...matching the passed in strings against a hash with values that are the classes or something.  And I don't love that idea, either!

lib/fake_issues.rb
```diff
@@ -0,0 +1,14 @@
+require "fake_api_response_wrapper"
+
+class FakeIssues
+  include Singleton
```
jduss4, line 4: This is a cool idea!

spec/models/api_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+  end
+end
+
+RSpec.describe Api do
+  describe "in a testing environment" do
+    describe "#self.request" do
+      it "returns the test request" do
+        result = Api.request("testsystem", "endpoint")
+        expect(result).to eq "test request"
```
jduss4, line 31: this is a cool test

jduss4 (commented):
(sorry, have to stop reviewing for a meeting, so just publishing what I have so far!  It's an interesting solution and I think could be really helpful moving forward!)

app/models/api.rb
Sgtpluck: i think because there are so many paths (fakedata vs no, system, endpoint) the hash would not be amazing, but i am happy to spec it out if it feels more readable! (or an alternative)

Sgtpluck: does this feel any better?

```
class Api
  def self.request(system, endpoint, args = nil)
    Api
      .namespace
      .const_get(system.capitalize)
      .const_get(endpoint.capitalize)
	  .new(*args)
	  .request
  end

  def self.namespace
    if Rails.env == "development" || Rails.env == "test"
      Api::FakeData
    else
      Api
    end
  end
end
```

rahearn: I think I like this version better. It'd likely also make Brakeman happier.

app/models/api/fake_data/hses.rb
rahearn: [suggestion] how about

```
def initialize(options = {})
  @id = options[:id]
end
```

I think passing an options hash around explicitly is a bit cleaner, though I guess you'd also need an initializer for Issues. I think I prefer that form, and there could be a common base class that provides the no-op initializer for all of the FakeData classes.

spec/models/api/fake_data/hses_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+require "fake_issues"
+
+RSpec.describe Api::FakeData::Hses do
+  describe "Issue" do
+    let(:complaint) { FakeIssues.instance.json[:data][4] }
+
+    describe "#request" do
+      let(:issue_endpoint) { Api::FakeData::Hses::Issue.new([:id, complaint[:id]]) }
```
rahearn, line 9: [nitpick] I think this array format is difficult to read and reason about, and so I think that readability outweighs having to put an initializer on `Issues`

spec/models/api_spec.rb
rahearn: +1

rahearn (commented):
I really love the structure of this solution.

app/models/api.rb
Sgtpluck: I will add a comment! what do you two think about having an endpoint validity check as well? would that be good for security? or is it Too Much(tm)

```
SYSTEMS = ["hses"]
ENDPOINTS = ["issue", "issues"]

def self.valid_request?(system, endpoint)
  SYSTEMS.include?(system) && ENDPOINTS.include?(endpoint)
end
```

Sgtpluck: also this is a lot of methods now, some which take arguments. should i instantiate an instance?

rahearn: I think that would be a very good thing to do if we were allowing users to specify the system or endpoint. Since they're hard-coded in the controllers we should get failures of `const_get` when running unit tests if bad values are passed so I don't think it adds much.

app/models/api/fake_data/hses.rb
Sgtpluck: it's definitely cleaner!

i was thinking about that no-op initializer! for the moment i could just put it in `Hses` maybe?

app/models/api.rb
rahearn: > also this is a lot of methods now, some which take arguments. should i instantiate an instance?

I don't think so, especially as long as the arguments stay on `self.request`.

Oh, also - we could avoid the `options` issue if we move that from the initializer to the request method.

Issue #87, opened 2021-08-23
Task: Terraform for Postgres


jduss4:
Task
Set up RDF for Rails app in staging

Part of story https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/90

Acceptance:
Given that I am a developer
When I run terraform locally
Then I should have access to a database from the Rails app in cloud.gov

PR #89, opened 2021-08-24
Terraform for RDS

jduss4:
## Description of change

Terraform code for creating RDS instance that will bind to the Rails application

## Acceptance Criteria

- [ ] creates RDS instance
- [ ] customizable for multiple environments
- [ ] well organized Terraform code, everybody is happy
- [ ] no secret / passwords rattling around in the world at large

## How to test

Otherwise, follow the instructions in `terraform/README.md` to make sure that Terraform is running correctly and can set up RDS in your sandbox.  You won't actually be able to bind anything to your application unless if you also set up rails running in your cloud.gov sandbox....I did not write down those instructions but can add them here if that would be helpful.

## Issue(s)

* closes https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/87

## Checklist

- [x] Meets issue criteria
- [x] Project board status updated

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


terraform/shared/database/main.tf
```diff
@@ -0,0 +1,23 @@
...
+
+###
+# RDS instance
+###
+
+data "cloudfoundry_service" "rds" {
+  name = "aws-rds"
+}
+
+resource "cloudfoundry_service_instance" "rds" {
```
Sgtpluck, line 18: [musing] i think based on the above documentation this is supposed to create a service instance if it doesn't exist. did you try it without first creating the service instance manually? sorry if this is a silly question!

terraform/README.md
```diff
@@ -0,0 +1,59 @@
+# Terraform
+
+## Get started
```
Sgtpluck, line 3: [suggestion/blocking] this "getting started" is only for a local setup, which I don't think we currently need, as we are running our dev environments locally with `rails s`

I think it would be helpful to start with the structure of the tf files, and have the local setup below, since that's more informative than anything else. 

also, how do we run it for staging/production? (i personally think i'd like to make it into a GitHub action ultimately so we can automate it as much as possible.)

terraform/database_dev/providers.tf
```diff
@@ -0,0 +1,30 @@
...
+  required_providers {
+    cloudfoundry = {
+      source  = "cloudfoundry-community/cloudfoundry"
+      version = "0.12.6"
+    }
+    aws = {
+      version = "~> 3.11.0"
+    }
+  }
+  # backend "s3" {
```
Sgtpluck, line 12: [suggestion] i don't think s3 is necessary! we could axe this. (we should anyways if it's commented out)

terraform/README.md
```diff
@@ -0,0 +1,59 @@
...
+    cf bind-service < YOUR-APP > < YOUR-RDS-SERVICE >
+    cf restage < YOUR-APP >
+    ```
+
+## Structure
+
+Each environment has its own module, which relies on a shared module for everything except the providers code and environment specific variables and settings.
+
+```
+- database_<env>/
```
Sgtpluck, line 44: [question] if we have different services, are we planning on splitting them into their own module? or having them all under `/dev`? i am just wondering if we need the `database_` namespacing.

terraform/database_dev/variables.tf
```diff
@@ -0,0 +1,35 @@
+variable "cf_api_url" {
```
Sgtpluck, line 1: [suggestion/blocking] i think we can definitely take this out, as it's the same in the `shared` variable file.

[question] actually, can we take out all the variable blocks here? i think we just need the definitions of the variables if they are different than what's in the shared file. 

IE:
`"env" = "dev"`

Sgtpluck (changes requested):

terraform/README.md
```diff
@@ -0,0 +1,59 @@
...
+    ```
+1. Run terraform with
+    ```bash
+    terraform init
+    terraform plan
+    terraform apply
+    ```
+1. Assuming all is well, now you can bind RDS to your app
+    ```bash
+    cf bind-service < YOUR-APP > < YOUR-RDS-SERVICE >
```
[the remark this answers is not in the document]
jduss4: Yeah, I looked at that, but I wasn't sure if route binding was the same as binding to an app?  I can definitely look into it more and see if I can figure out what a route is in this context :)

terraform/shared/database/main.tf
jduss4: It seems to work for me when I destroy all the services!  In fact, it complains rather bitterly if it didn't know a service existed when it starts to run for the first time, as I learned when I had manually created an rds instance to make sure I understood how to, then tried to run this terraform code.  I don't think it's a silly question!

terraform/shared/database/variables.tf
```diff
@@ -0,0 +1,41 @@
...
+  default     = "hhs-acf-ohs-hses"
+}
+
+variable "cf_password" {
+  type        = string
+  description = "secret; cloud.gov deployer account password"
+  sensitive   = true
+}
+
+variable "cf_space_name" {
```
rahearn, line 25: [issue] This variable (or at least it's default value) should be in the `stage` directory and not the shared one.

terraform/database_dev/providers.tf
jduss4: I left it in because I wanted to talk about it, so yay, we're talking about it!

I don't know if it's necessary or not.  From my understanding terraform needs to store its state for each environment somewhere it can refer to on subsequent runs. Can GitHub Actions accomplish that?  If not, I think we need to have an s3 bucket that's holding the state.  Since I was just developing locally for a sandbox, I was writing the state locally, but I didn't want to totally take out references to remote state storage, either, so we could figure out what we want to do about it!

terraform/README.md
jduss4: I would be happy to organize this however we want.  I was taking inspiration from the ECLKC vendor team, which organized them like `service_env` and then `modules/service`, but I think it makes sense to have `env/service` instead.

terraform/database_dev/variables.tf
jduss4: `env = dev` would be super cool, and take away my feelings of ickiness at basically having to duplicate the variables file.  I will do some experimentation!

terraform/README.md
jduss4: I guess it depends on whether we have to run Terraform ourselves or whether we can automate that via GH Actions.  I do not know how to run staging / production other than running it locally, but I can look into automating that, too, in which case you're right, we wouldn't really need the getting started stuff anymore.

jduss4: I should also mention that a lot of the getting started instructions were notes I was taking for myself as I went, to make sure I knew what I had done on my fantastic voyage.

rahearn (commented):
I think you replied to some things while this was still in draft, so things might not make a ton of sense.

Sgtpluck (approved):

terraform/shared/database/main.tf
```diff
@@ -0,0 +1,23 @@
...
+  name             = "ohs-ct-rds-${var.env}"
+  space            = data.cloudfoundry_space.space.id
+  service_plan     = data.cloudfoundry_service.rds.service_plans["micro-psql"]
```
rahearn, line 21: [thought] This is good for now, but we should make a note to potentially pass a variable for the plan type once we add a production environment.

rahearn (approved):
🎉

Issue #92, opened 2021-08-26
Task: Terraform via GitHub Actions


jduss4:
This task is part of #90 and blocked by https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/91

Full User Story:
As a developer
I want to manage services through code
So that we can document and automate our infrastructure.

This task is scoped to:
- [] set up GitHub Actions to automatically deploy any changes to Terraform code

PR #109, opened 2021-09-08
Terraform actions

jduss4:
## Description of change

Creates 2 GitHub Action jobs:
- runs `terraform plan` when a PR to the main branch is opened, and outputs the results of `terraform plan` to a comment. Should fail if `terraform plan` fails
- runs `terraform apply` when changes to the terraform staging and shared directories are made and pushed to the main branch

At least I hope that's what happens, because I don't know how to test the second one.

## Acceptance Criteria

- `terraform plan` outputs content to a PR issue
- `terraform apply` runs successfully on main branch (not sure how to test this?)

## How to test

Great question!  I don't recommend the `act` library, unfortunately. I've tested manually here, so hopefully this PR getting a successful comment is enough evidence for the first goal of this PR.

## Issue(s)

* closes https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/92

## Checklist

- [X] Meets issue criteria
- [NO] Code is meaningfully tested
- ~[ ] Meets accessibility standards (WCAG 2.1 Levels A, AA)~

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/.github/workflows/deploy-stage.yml
+++ b/.github/workflows/deploy-stage.yml
@@ -22,8 +22,8 @@ jobs:
         uses: tj-actions/changed-files@v1.1.2
         with:
           files: |
-            terraform/shared/**
-            terraform/stage/**
+            terraform/shared
+            terraform/stage
 
       - name: Terraform apply
         if: steps.changed-files.outputs.any_changed == 'true'
--- a/.github/workflows/deploy-stage.yml
+++ b/.github/workflows/deploy-stage.yml
@@ -17,18 +17,6 @@ jobs:
         with:
           fetch-depth: 2
 
-      - name: Deploy app
-        uses: 18F/cg-deploy-action@main
-        env:
-          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
-          HSES_AUTH_CLIENT_SECRET: ${{ secrets.HSES_AUTH_CLIENT_SECRET }}
-        with:
-          cf_username: ${{ secrets.CF_USERNAME }}
-          cf_password: ${{ secrets.CF_PASSWORD }}
-          cf_org: hhs-acf-ohs-hses
-          cf_space: ct-stage
-          push_arguments: "--vars-file config/deployment/stage.yml --var rails_master_key=$RAILS_MASTER_KEY --var hses_auth_client_secret=$HSES_AUTH_CLIENT_SECRET"
-
       - name: Check for changes to Terraform
         id: changed-files
         uses: tj-actions/changed-files@v1.1.2
@@ -46,3 +34,15 @@ jobs:
           TF_VAR_cf_user: ${{ secrets.CF_USERNAME }}
           TF_VAR_cf_password: ${{ secrets.CF_PASSWORD }}
         run: terraform apply -auto-approve -input=false
+
+      - name: Deploy app
+        uses: 18F/cg-deploy-action@main
+        env:
+          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
+          HSES_AUTH_CLIENT_SECRET: ${{ secrets.HSES_AUTH_CLIENT_SECRET }}
+        with:
+          cf_username: ${{ secrets.CF_USERNAME }}
+          cf_password: ${{ secrets.CF_PASSWORD }}
+          cf_org: hhs-acf-ohs-hses
+          cf_space: ct-stage
+          push_arguments: "--vars-file config/deployment/stage.yml --var rails_master_key=$RAILS_MASTER_KEY --var hses_auth_client_secret=$HSES_AUTH_CLIENT_SECRET"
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-stage.yml
```diff
@@ -26,3 +28,21 @@ jobs:
           cf_org: hhs-acf-ohs-hses
           cf_space: ct-stage
           push_arguments: "--vars-file config/deployment/stage.yml --var rails_master_key=$RAILS_MASTER_KEY --var hses_auth_client_secret=$HSES_AUTH_CLIENT_SECRET"
+
+      - name: Check for changes to Terraform
+        id: changed-files
+        uses: tj-actions/changed-files@v1.1.2
+        with:
+          files: |
+            terraform/shared/**
```
jduss4, line 37: [comment] I am not actually sure if this syntax will work, as I didn't see a lot of guidance about directories in the [changed-files documentation](https://github.com/tj-actions/changed-files), but I hoping it will?

.github/workflows/deploy-stage.yml
```diff
@@ -26,3 +28,21 @@ jobs:
           cf_org: hhs-acf-ohs-hses
           cf_space: ct-stage
           push_arguments: "--vars-file config/deployment/stage.yml --var rails_master_key=$RAILS_MASTER_KEY --var hses_auth_client_secret=$HSES_AUTH_CLIENT_SECRET"
+
+      - name: Check for changes to Terraform
```
jduss4, line 32: [comment] Thanks for the code that I lifted straight out of your PR, @rahearn !

.github/workflows/terraform-stage.yml
```diff
@@ -0,0 +1,79 @@
...
+
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: output
+            })
+
+      - name: Terraform plan status
+        if: ${{ always() }} && steps.plan.outcome == 'failure'
```
jduss4, line 78: [question] This is currently sending an exit code 1, and I kinda wonder if it's because `always()` doesn't play well with && ? This is currently the way it is because I'm not using `continue-on-error` during the `terraform plan` step, although I think I would prefer that. I took it out after the comment someone made that `continue-on-error` seemed like code smell, but I think adding `always()` to the following actions is not my favorite, either.
Thoughts?

.github/workflows/deploy-stage.yml
rahearn: [suggestion] I don't think it will work, but just `terraform/shared` and `terraform/stage` will.

Basically what it's doing with these is running `git diff --name-only | grep -E FILE_NAME` to filter the files changed down to the name. I did a quick and dirty test and a file with `**` at the end had this error: `grep: repetition-operator operand invalid`

.github/workflows/terraform-stage.yml
```diff
@@ -0,0 +1,79 @@
...
+        id: validation
+        run: terraform validate -no-color
+
+      - name: Terraform plan
+        id: plan
+        run: terraform plan -no-color -input=false 2>&1 | tee plan_output.txt
+
+      - name: Read Terraform plan output file
+        id: terraform_output
+        uses: juliangruber/read-file-action@v1
```
rahearn, line 44: [suggestion] this needs an `if: ${{ always() }}` to read in the plan output in the case of failure.

.github/workflows/terraform-stage.yml
rahearn: I think the closing brackets around `always()` are tripping it up. If left in, this should either be

`if: ${{ always() && steps.plan.outcome == 'failure' }}` or `if: always() && steps.plan.outcome == 'failure'` or maybe even just `if: steps.plan.outcome == 'failure'`

[suggestion] this whole step can be removed. The job will fail properly when plan fails since you removed `continue-on-error`

terraform/stage/providers.tf
```diff
@@ -15,6 +15,5 @@ terraform {
     key     = "terraform.tfstate.stage"
     encrypt = "true"
     region  = "us-gov-west-1"
-    profile = "ct-terraform"
```
rahearn, line 18: [chore] If this needs to be removed, then we should update the README for running locally to either not put the credentials in that profile (rather keep that) or add a step to locally set the proper ENV variable to access the credentials where they're stored (I think it's `AWS_PROFILE` but double check that)

rahearn (changes requested):

.github/workflows/terraform-stage.yml
jduss4: Oh yeah, I bet you're right :)  Thanks!

terraform/stage/providers.tf
jduss4: I do believe it needs to be removed, so I can make those changes.

rahearn (approved):
🎉

Issue #131, opened 2021-09-14
Task: Incorporate real HSES API response


rahearn:
Task: implement basic connection to HSES API
Related to story https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/133

- [x] utilizing real HSES API response in staging

jduss4:
Closed with #132

PR #132, opened 2021-09-15
Connect to hses

jduss4:
## Description of change

Adds a very basic connection to HSES staging API using basic auth. We will need to change this substantially, but as a proof of concept, ✅  -- yay it works

## Acceptance Criteria

- Connects to HSES staging API, retrieves data to display on `/complaints` page.
- username / password are secure

Things I think we should worry about in the next PRs:
- parameters
- pagination
- errors
- rethinking the structure of our classes

## How to test

You will need to disable the fake data one way or another. What I was doing was just going to `models/api.rb` and swapping out `self.needs_fake_data?`

## Issue(s)

Issue https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/131 is ostensibly the one at hand, but @Sgtpluck and I had a discussion about whether or not to add staging data in development which ended up at "maybe not," therefore I'm not sure this actually addresses the entirety of that issue.

## Checklist

- [ ] Meets issue criteria
- [ eh ] Code is meaningfully tested
- [ not yet ] Boundary diagram updated

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


app/models/api/hses.rb
```diff
@@ -18,8 +18,17 @@ def request
   end
 
   class Issues
+    def initialize(user:)
+      @username = user["uid"]
+    end
+
     def request
-      @issues ||= FakeIssues.instance.json
+      res = Api.query_api(
```
rahearn, line 26: [question] Why call back up to the `Api` level for this. We know we'll have multiple API endpoints so might as well keep that separation now.

config/environments/production.rb
```diff
@@ -118,6 +118,12 @@
   # config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
 
+  # HSES API
+  config.x.hses.api = ENV["HSES_API"]
```
rahearn, line 122: [suggestion] I don't think this env var name is quite specific enough. What about `HSES_API_HOSTNAME`?

jduss4:
I ran pa11y locally with `RAILS_ENV=ci` and I wasn't able to replicate the issue above

rahearn:
> I ran pa11y locally with `RAILS_ENV=ci` and I wasn't able to replicate the issue above

I'm able to reproduce locally, looks like missing config for the HSES api is at fault.

jduss4:
Huh, wonder how I managed to set that up incorrectly. I think it's inheriting from production, which I have set to environment variables but they aren't in Github yet.  I can stick them in for the staging environment, hopefully that will at least help the github action out!

app/models/api/hses.rb
jduss4: The behavior of sending and receiving a response felt like fairly generic behavior. I realize that the authentication methods may vary, but it still feels like there ought to be some central class that handles it. Then maybe we have a subclass for the different systems that override some of the implementation but continue to rely on the same error handling / etc.  I dunno, I don't really want to move it back because it felt very weird in the `Issues` class....maybe I can find a home in an Hses-specific class that would make me feel better.

Sgtpluck: [comment/suggestion] i think it should be on the Api class -- but i think we should reorganize this a bit to have Api just hold the http behavior and the individual systems hold their own config information. what i am envisioning right now is this: 

if we have the classes inherit from Api then they will get the classes from them. so we can add 
`class Api::Hses < Api` (is it weird to namespace AND inherit from there? i say probably, so maybe we want to have one or both be different?) and then we could have thinks like 

```ruby
class Api::Hses < Api
  def host
    Rails.configuration.x.hses.api_base
  end
  def path
  end
  def query
  end
end
```

and then in Api we could have 
```ruby
class Api
...
  def query_api
    uri = build_uri(host, path, query)
    res = send_api_request(uri)
    body = format_response_body(res.body)
    {code: res.code, body: body}
  end
```

so this line could just be 
`res = query_api`

or we could even have `res` be its own method so:

```ruby
def request
  {code: res.code, body: format_response_body(res.body)}
end

def res
  @res ||= query_api
end
```

i like this because then we don't have to keep track in Api or somewhere what the configuration of each system is -- the system gets to keep track of it. and if there are different queries for specific endpoints they can override those by defining their own `query` method!

what do you all think?

jduss4: Totally agree.  I was thinking of something along the same lines.  Basically:

- ApiDelegator: has logic which helps you get to the correct type of Api subclass (Hses / FakeData::Hses / Tta, etc)
- Api:  has defaults like you're describing, including stuff like logic for pagination and sorting
- Api subclasses:  have the specific implementation details and overrides

That makes wayyy more sense to me.  Want me to implement that on this PR?

Sgtpluck: oh yeah i like this structure! i would love to see it on this PR!

jduss4:
I made some structural changes:

ApiDelegator pulls logic out of Api which is used to determine which particular type of class to create based on the system and reality (aka: HSES and fake data, or HSES and real data)

ApiRequest has methods for actually sending requests to a service and could be used in the future to hold logic for pagination, sorting, etc. It's currently doing a little bit of response work that might be better off as an ApiResponse class, but nothing too major right now.

Now, Api::Hses::Issue and Api::Hses::Issues both inherit from ApiRequest, which means that they only need to override a couple settings via methods and then they are good to go.

app/controllers/complaints_controller.rb
```diff
@@ -1,20 +1,29 @@
+require "api_delegator"
+
 class ComplaintsController < ApplicationController
   before_action :check_pa11y_id, only: :show, if: -> { Rails.env.ci? }
 
   def index
-    @complaints = Api.request("hses", "issues")[:data]
+    api = ApiDelegator.use("hses", "issues", {user: session["user"]})
+    res = api.request
+    @complaints = res[:data]
```
Sgtpluck, line 9: [nit] this could be a single line 
`@complaints = api.request[:data]`

app/controllers/complaints_controller.rb
```diff
@@ -1,20 +1,29 @@
...
   end
 
   def show
-    @complaint = Complaint.new(Api.request("hses", "issue", {id: params[:id]})[:data])
+    @complaint = Complaint.new(
```
Sgtpluck, line 13: can't wait to just have a complaint endpoint 😭

app/models/api/hses.rb
```diff
@@ -1,25 +1,49 @@
+require "api_request"
 require "fake_api_response_wrapper"
 require "fake_issues"
 
-class Api::Hses
-  class Issue
-    include FakeApiResponseWrapper
-    attr_accessor :id
+module Api::Hses
```
Sgtpluck, line 5: [thought] i think if we keep this a class we can have it inherit from ApiRequest so every subclass doesn't need to. thoughts?

app/models/api/hses.rb
```diff
@@ -1,25 +1,49 @@
...
+  end
+end
+
+class Api::Hses::Issues < ApiRequest
+  def initialize(user:)
+    @username = user["uid"]
+  end
+
+  def request
+    res = query("username=#{@username}&types=1")
```
Sgtpluck, line 36: [suggestion] hmmm i don't like that we're passing through something to this method.

it would make more sense to me i think for the current `query` method to be renamed to something like 
`response` and then in this class have

```ruby
def query
  "username=#{@username}&types=1"
end
```

app/models/api_request.rb
```diff
@@ -0,0 +1,47 @@
+class ApiRequest
+  def query(query)
+    uri = request_uri(query)
```
Sgtpluck, line 3: if you take my above suggestion this can become 

```ruby
def response
  res = send_api_request(request_uri)
  format_response(res)
end
```
(i would do that all in one line and not create the var but that's my personal style :))

app/models/api_request.rb
```diff
@@ -0,0 +1,47 @@
...
+    res = send_api_request(uri)
+    format_response(res)
+  end
+
+  def request_uri(query = nil)
+    URI::HTTPS.build(
+      host: host,
+      path: path,
+      query: query
```
Sgtpluck, line 12: this method won't need a param (it will be coming from the class itself) if you take my above suggestion!

app/models/api_request.rb
```diff
@@ -0,0 +1,47 @@
...
+  def format_response(res)
+    {
+      code: res.code,
+      body: JSON.parse(res.body)
+    }.with_indifferent_access
+  end
+
+  # Inheriting class defines
+  def host
+    ""
```
Sgtpluck, line 27: [suggestion] should we have this and path throw an error if they're not implemented in the specific classes?

app/models/api_request.rb
```diff
@@ -0,0 +1,47 @@
...
+  # Inheriting class defines
+  def path
+    ""
+  end
+
+  # NOTE: this will need refactoring if we connect to multiple APIs
+  def send_api_request(uri)
+    req = Net::HTTP::Get.new(uri)
+    req.basic_auth(
+      Rails.configuration.x.hses.api_username,
```
Sgtpluck, line 39: [suggestion]if you make these methods then it's future proof, but also fine if that feels like an overoptimization now!

Sgtpluck (commented):
I didn't have time to totally finish but here are some comments for you to think about!

app/models/api/hses.rb
jduss4: I agree that would be nice, but I was having trouble figuring out how to keep something namespaced while also having inheritance, since those aren't the same thing.

Like, `Api::Hses` is different than `Hses < Api` but I don't really want to say `Api::Hses < Api` ... although I might just be missing something about how to have my cake and inherit some, too.

app/models/api/hses.rb
jduss4: Yeah that's probably cleaner, and then we can revisit once we start getting things actually passed through from parameters the user is selecting.

app/models/api/hses.rb
Sgtpluck: But it would be `Api::Hses < ApiRequest` which seems okay to me

jduss4: As far as I'm aware there aren't really "nested" classes in Ruby, so I don't think you can inherit methods through namespacing / nesting. I tried this out just now to check, and didn't get the error I was expecting so maybe there a way to access the methods after all, but I will need to look into it more:

```ruby
class ApiRequest
  def test
    puts "from ApiRequest"
  end
end

class Api
  class Hses < ApiRequest
    class Issue
    end
  end
end

ApiRequest.new.test
Api::Hses::Issue.new.test
```

> from ApiRequest
> testing.rb:18:in `<main>': private method `test' called for #<Api::Hses::Issue:0x00007fd55f8d39e8> (NoMethodError)

app/models/api_request.rb
jduss4: I like that idea!  I'll do it for query, too.

app/models/api/hses.rb
```diff
@@ -1,25 +1,53 @@
...
+class Api::Hses::Issues < ApiRequest
+  def initialize(user:)
+    @username = user["uid"]
+  end
 
-    def initialize(id:)
-      @id = id
-    end
+  def request
+    res = response
```
jduss4, line 36: I don't like this line very much, so I'm happy to take some suggestions on what to do about this.  I think what I would like, ideally, is that `response` isn't just a hash, but is an instance of an ApiResponse class with things like `response.code`, `response.data`, `response.current_page` and that type of thing.  For now, this is enough to get us up and running, I think.

app/models/api/hses.rb
Sgtpluck: eh ok, no biggie

app/models/api/hses.rb
Sgtpluck: since we're creating a new Api::Hses::Issues instance each time we call this, can we memoize `response` in ApiRequest?

app/models/api/hses.rb
```diff
@@ -1,25 +1,52 @@
...
+  def request
+    details_wrapper.merge(
+      data: FakeIssues.instance.json[:data].sample
+    )
+  end
+
+  private
+
+  def host
+    Rails.configuration.x.hses.api_base
```
Sgtpluck, line 26: can this go in Api::Hses so it doesn't have to be repeated?

app/models/api_request.rb
```diff
@@ -0,0 +1,55 @@
+class ApiRequest
+  def request_uri
```
Sgtpluck, line 2: [question] should this be a private method?

spec/models/api_request_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        allow_any_instance_of(ApiRequest).to receive(:path).and_return "/path"
+        allow_any_instance_of(ApiRequest).to receive(:query).and_return "?some=query"
+
+        res = ApiRequest.new.response
+        expect(res).to match({code: "200", body: {"meta" => {}, "data" => []}})
+      end
+
+      it "is memoized" do
+        api = ApiRequest.new
+        expect(api).to receive(:get_response).once.and_return rand
```
Sgtpluck, line 26: what's rand?

app/models/api/hses.rb
jduss4: I think this is our nested classes problem again :(  Maybe I can figure out some way of importing the methods from another class without inheriting them.

spec/models/api_request_spec.rb
jduss4: This made more sense when it was `rand()` but our linter didn't like the parentheses.  I just made a random number to match up later, but also I can drop the comparison and just rely on the `.once` part instead, easily enough.

Sgtpluck: maybe have it return a hast that is like:
{ random_test_response": rand } or something to make it clearer what's happening?

app/models/api/hses.rb
jduss4: I added it as a method on the module Api::Hses and included the module directly, which has done the trick, even if not particularly elegantly!

spec/models/api_request_spec.rb
rahearn: I like @Sgtpluck's idea, and also, there's a chance that this will return the same random value no matter how many times it's called. You likely need to put that into a block for it to be re-evaluated each time `get_response` is called.

jduss4: good points!  I will do that!

spec/models/api_request_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        allow_any_instance_of(ApiRequest).to receive(:query).and_return "?some=query"
+
+        res = ApiRequest.new.response
+        expect(res).to match({code: "200", body: {"meta" => {}, "data" => []}})
+      end
+
+      it "is memoized" do
+        api = ApiRequest.new
+        expect(api).to receive(:get_response).once.and_return "value"
+        api.response
```
Sgtpluck, line 27: [thought] i think i would prefer this to be written as 
`expect(api.response).to eq api.response` since that is an expectation as well as calling the method twice. (but this is more of a nitpick i guess? not blocking for sure)

Sgtpluck (approved):
A couple of minor thoughts/suggestions. All nonblocking so I'm hitting approve.

lib/api_delegator.rb
```diff
@@ -1,22 +1,21 @@
-class Api
-  def self.request(system, endpoint, args = {})
-    Api
+module ApiDelegator
```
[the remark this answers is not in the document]
rahearn: [praise] I love the change from `Api` to `ApiDelegator`

manifest.yml
```diff
@@ -9,6 +9,9 @@ applications:
     HSES_AUTH_BASE: ((hses_auth_base))
     HSES_AUTH_CLIENT_ID: ((hses_auth_client_id))
     HSES_AUTH_CLIENT_SECRET: ((hses_auth_client_secret))
+    HSES_API_BASE: ((hses_api_base))
+    HSES_API_USERNAME: ((hses_api_username))
+    HSES_API_PASSWORD: ((hses_api_password))
```
rahearn, line 14: [suggestion] Add lines to `config/deployment/stage.yml` for `hses_api_base` and `hses_api_username` and the appropriate env var and `push_arguments` to `.github/workflows/deploy-stage.yml` for the password to be grabbed from the staging secrets

lib/api_delegator.rb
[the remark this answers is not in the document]
jduss4: Well, since it doesn't behave like a class, we're not going to need instance methods, and it's not being inherited by anything, it made sense to me to just collect a couple methods up in a module.  I don't feel terribly strongly about it, though, so if you would prefer a class I can change that.

manifest.yml
jduss4: Good catch. Guess I gotta delete a couple base / username secrets that I stuck into github when I got too secret happy.

lib/api_delegator.rb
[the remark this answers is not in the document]
rahearn: Nah, that all sounds like a good reason to make it a module.

rahearn (approved):
🎉

PR #143, opened 2021-09-17
Manually link TTA report and complaint in tracker

### LINKED ISSUE #142: Link existing TTA Activity Report to Complaint and display detail
User Story:
As a program specialist
I want to link an existing activity report ID to a complaint
So that I can track progress that was kicked off out-of-band

Acceptance:
When I add an AR ID to the complaints details
Then I can see that AR in the summary view

This is the first step of #96

Sgtpluck:
## Description of change
This change adds a form to link a TTA activity report and a complaint.
It includes:
* opening/closing a create form with javascript.
* Creates a link between activity report and complaint
* Adds Report Id in summary box.

It does not include:

* a real API call
* error validation
* editing/deleting a TTA report
* adding TTA details on the timeline

## Acceptance Criteria
When I add an AR ID to the complaints details
Then I can see that AR in the summary view

## How to test
- [ ] go to a complaint detail
- [ ] there should be an "Add activity report" link
- [ ] click link
- [ ] an "add" activity report form should open
- [ ] click cancel
- [ ] form should go away
- [ ] add any string
- [ ] page will reload with a new section on the summary that says "TTA activity" and have that string in it.
- [ ] reload the page
- [ ] TTA activity section should still be there

Screen Shot 2021-09-17 at 11 40 01 AM
## Issue(s)

* closes #142

## Checklist

- [x] Meets issue criteria

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


app/models/api.rb
```diff
@@ -1,7 +1,7 @@
 class Api
   def self.request(system, endpoint, args = {})
     Api
-      .namespaced_class(system.capitalize, endpoint.capitalize)
+      .namespaced_class(system.camelize, endpoint.camelize)
```
jduss4, line 4: [praise] good call!

app/views/complaints/_tta_activity_report_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="ct-section-divider">
+  <p class="display-flex flex-justify">
+    <label for="tta-report-id">Activity Report ID</label>
+    <span><%= link_to "Cancel", "#", id: "js-close-tta-form" %></span>
+    <div>
+      <%= form_with url: "/issue_tta_reports", method: "post", class: "usa-search display-flex" do %>
```
jduss4, line 6: [suggestion] Probably because of my background at a place where we deployed quite frequently to sub-uris, a url like `/issue_tta_reports` makes me a worried. Could this instead be `url: new_issue_tta_reports` instead?

app/views/complaints/show.html.erb
```diff
@@ -7,17 +7,35 @@
...
-        </p>
+        <div class="ct-section-divider">
+          <h2>Summary</h2>
+          <p>
+            <strong><%= @complaint.grantee %>:</strong>
+            <%= @complaint.summary %>
+          </p>
+        </div>
+        <% unless @tta_reports.empty? %>
+          <%= render "tta_report_list", :tta_reports => @tta_reports %>
```
jduss4, line 18: [suggestion / nitpick] if you want to simplify this when you're using symbols, you can write `tta_reports: @tta_reports` rather than use a rocket.  We tried to avoid symbols + rockets at my old work but I mean, they work fine so I guess do whatever you want with this suggestion :)

app/views/complaints/show.html.erb
```diff
@@ -7,17 +7,35 @@
...
+        </div>
+        <% unless @tta_reports.empty? %>
+          <%= render "tta_report_list", :tta_reports => @tta_reports %> 
+        <% end %>
+        <div class="display-none" id="tta-activity-form">
+          <%= render "tta_activity_report_form", :complaint => @complaint %>
+        </div>
+        <div class="ct-section-divider">
+          <p>
+            <span class="text-middle">
```
jduss4, line 25: [astonishment] I am unfamiliar with this way of doing icons, whoa

app/views/complaints/show.html.erb
```diff
@@ -7,17 +7,35 @@
...
+              <svg class="usa-icon text-primary" aria-hidden="true" focusable="false" role="img">
+                <use xlink:href="/assets/img/sprite.svg#add_circle"></use>
+              </svg>
+            </span>
+              <%= link_to "Add TTA Activity Report", "#", class: "ct-add-report-link", id: "js-add-tta-report" %>
```
jduss4, line 30: [question] Hmm, I don't love that this is a link that behaves more like a button. Do you think we should add any sort of aria indication so that it's clear that this isn't actually taking the user to a new page to add the activity report?

app/views/complaints/show.html.erb
```diff
@@ -7,17 +7,35 @@
...
-          <%= @complaint.summary %>
-        </p>
+        <div class="ct-section-divider">
+          <h2>Summary</h2>
+          <p>
+            <strong><%= @complaint.grantee %>:</strong>
+            <%= @complaint.summary %>
+          </p>
+        </div>
+        <% unless @tta_reports.empty? %>
```
jduss4, line 17: [suggestion] I think that we could use `<% if @tta_reports.present? %>` here instead, which to me seems more intuitive than "unless empty"

app/javascript/packs/tta-report.js
```diff
@@ -0,0 +1,11 @@
+document.querySelector("#js-add-tta-report").addEventListener('click', event => {
+  const form = document.querySelector("#tta-activity-form")
+  form.classList.remove("display-none")
```
jduss4, line 3: [suggestion] could we automatically focus on the new form when this is opened?

jduss4 (commented):
This is looking great!  This is a complicated little doodad, you are right!

app/views/complaints/show.html.erb
Sgtpluck: we can make it a button ... but i made it a link because i also don't think it's acting really like a button (it's not posting anything.) it feels more like an on-page anchor link. (if i take your suggestion and focus on the input, it would feel more like that too i think).

i am very open to suggestions on what element to use to better indicate what's happening here to screen readers!

app/javascript/packs/tta-report.js
Sgtpluck: heck yeah we can

app/views/complaints/show.html.erb
Sgtpluck: oookay i did some more research and i do think link is correct here (it takes us to a different part of the page.) i did make it better but changing the href to the id of the form input, which also means it focuses without using javascript, so this seems like the right way to do it!

Sgtpluck: okay having done more research and talked it over with @jduss4 i am reversing my thought -- since it shows/hides, it seems to be a button. i am going to add a button role to make sure this is read correctly by a screen reader and do a manual test of it.

app/javascript/packs/tta-report.js
```diff
@@ -0,0 +1,35 @@
...
+
+const closeForm = function () {
+  const input = document.querySelector("#tta-report-id")
+  input.value = ""
+  const form = document.querySelector("#tta-activity-form")
+  form.classList.add("display-none")
+}
+
+document.querySelector("#js-add-tta-report").addEventListener("keydown", event => {
+  if (event.keyCode == 32) {
```
Sgtpluck, line 14: as per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#accessibility_concerns

```
Warning: Be careful when marking up links with the button role. Buttons are expected to be triggered using the Space or Enter key, while links are expected to be triggered using the Enter key. In other words, when links are used to behave like buttons, adding role="button" alone is not sufficient. It will also be necessary to add a key event handler that listens for the Space key in order to be consistent with native buttons.

```

app/views/complaints/_tta_activity_report_form.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="ct-section-divider">
+  <p class="display-flex flex-justify">
+    <span> Activity Report ID</span>
+    <span><%= link_to "Cancel", "#js-add-tta-report", id: "js-close-tta-form", role: "button" %></span>
+    <div>
+      <%= form_with url: issue_tta_reports_path, method: "post", class: "usa-search display-flex", id: "tta-report-form"  do %>
+        <%= label_tag "tta-report-id", "TTA report display id", id: "tta-report-label" %>
```
rahearn, line 7: [suggestion] Update the label to `TTA report display ID` so that VoiceOver pronounces each of the letters of I.D. instead of the end of `did`

app/views/complaints/show.html.erb
```diff
@@ -7,17 +7,40 @@
...
+          <%= render "tta_activity_report_form", complaint: @complaint %>
+        </div>
+        <div class="ct-section-divider">
+          <p>
+            <span class="text-middle">
+              <svg class="usa-icon text-primary" aria-hidden="true" focusable="false" role="img">
+                <use xlink:href="/assets/img/sprite.svg#add_circle"></use>
+              </svg>
+            </span>
+              <%= link_to "Add TTA Activity Report", "#tta-report-id", {
```
rahearn, line 30: [question] Is using `button_to` here to get the semantics of a button way more styling work than having to implement the custom keydown handler to get a link to respond to both Space and Enter?

app/views/complaints/_tta_activity_report_form.html.erb
```diff
@@ -0,0 +1,14 @@
...
+        <%= text_field_tag :tta_report_id, nil, id: "tta-report-id", "aria-describedBy": "tta-report-label" %>
```
rahearn, line 8: [thought] `aria-describedBy` works well. You can also accomplish the same thing by removing the CSS from the `label#tta-report-label` and adding the `usa-sr-only` class to it. I think `visibility: hidden` causes the label to be hidden from VO as well as from sighted users

rahearn (commented):
Overall, I love this. Really great job of not over complicating the JS in use.

app/models/issue_tta_report.rb
```diff
@@ -0,0 +1,4 @@
+class IssueTtaReport < ApplicationRecord
+  validates :issue_id, :tta_report_display_id, :tta_report_id, presence: true
+  validates :tta_report_id, numericality: {only_integer: true}
```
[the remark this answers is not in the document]
Sgtpluck: i don't feel strongly one way or another -- i was just mimicking TTA's data types. 

what do you mean by "safer" in this instance?

app/views/complaints/_tta_report_list.html.erb
```diff
@@ -0,0 +1,19 @@
...
+        <% tta_reports.each_with_index do |report, i| %>
+          <div>
+            <%= report.tta_report_display_id %>
+          </div>
+        <% end %>
+      </p>
+    </span>
+    <span class="display-none">
+      <svg class="usa-icon text-primary" aria-hidden="true" focusable="false" role="img">
+        <use xlink:href="/assets/img/sprite.svg#edit"></use>
```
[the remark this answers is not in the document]
Sgtpluck: oh, probably! this is part of USWDS. do you have any experience writing a path helper for that? otherwise i can research.

app/views/complaints/show.html.erb
Sgtpluck: no, i suppose not -- it just feels weird to use button_to and then have to prevent the default of what that functionality wants to do. neither option is great, i suppose, but let do a quick pass of making the button a button

app/views/complaints/_tta_report_list.html.erb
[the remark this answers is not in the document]
rahearn: So... if not for the `#edit` I'm pretty sure `image_pack_path("img/sprite.svg")` would work. I'd start there and mess with it

app/models/issue_tta_report.rb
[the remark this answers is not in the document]
rahearn: Safer in that a change from integers to UUID wouldn't need to result in a db migration for us (though it would still require a code change for the validation)

app/views/complaints/_tta_activity_report_form.html.erb
Sgtpluck: i did all my experimenting with the label before i hid it. when it was visible it was not triggering in the screenreader until i used `aria-describedBy`

app/views/complaints/show.html.erb
Sgtpluck: i think we need to add some more js to get the icon to be clickable as well ugh

app/views/complaints/_tta_activity_report_form.html.erb
Sgtpluck: i do prefer the `usa-sr-only` class to creating my own custom css, though, so i am going to do that

app/controllers/issue_tta_reports_controller.rb
```diff
@@ -0,0 +1,23 @@
+class IssueTtaReportsController < ApplicationController
+  def create
+    @issue_tta_report = IssueTtaReport.new(
+      issue_id: params[:issue_id],
+      tta_report_display_id: tta_report_id,
+      tta_report_id: activity_data[:id]
+    )
+
+    if @issue_tta_report.save
```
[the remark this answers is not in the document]
Sgtpluck: i'm happy to hear arguments about whether or not this is a thing we currently need! it seems a bit of an overoptimization at present but would love to know your thoughts more.

app/models/issue_tta_report.rb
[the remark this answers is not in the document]
Sgtpluck: I guess I would just be surprised if TTAHub changed the data type of the ActivityReport's id, but migrations are inherently riskier than changing validations. what do you think @jduss4?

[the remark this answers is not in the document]
rahearn: I would also be surprised by that, but weirder things have happened. A more likely possibility is way down the road when `integer` -> `bigint`

Either way, my rule of thumb is to use strings for most things that don't involve math, even if they happen to currently be numbers

app/controllers/issue_tta_reports_controller.rb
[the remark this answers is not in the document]
rahearn: I'd be surprised if we didn't need it eventually, but I think the question does point to it being a new user story as opposed to a task on this one.

app/models/issue_tta_report.rb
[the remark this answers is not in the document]
jduss4: I think that using a string is fine, especially if we're using validation, so that this will match the other identifier fields.  That being said, I don't have a problem with it being an integer if we're expecting an integer.  Migrating to bigint shouldn't be a big deal in the future, if it was needed.  My official vote is slight preference towards string but I don't mind integer.

app/views/complaints/_tta_report_list.html.erb
```diff
@@ -10,10 +10,5 @@
         <% end %>
       </p>
     </span>
-    <span class="display-none">
```
rahearn: Should there be a replacement for this?

Sgtpluck: no, that design is changing i think to make it easier to link/unlink individual activity reports. (laura and i discussed that last week)

rahearn (approved):
🎉

app/models/issue_tta_report.rb
```diff
@@ -0,0 +1,4 @@
+class IssueTtaReport < ApplicationRecord
+  validates :issue_id, :tta_report_display_id, :tta_report_id, presence: true
+  validates :tta_report_id, format: {with: /\A[0-9]*\z/, message: "only allows numbers"}
```
jduss4, line 3: [bafflement] TIL something new about regexes

app/views/complaints/_tta_report_list.html.erb
```diff
@@ -0,0 +1,14 @@
+ <div class="ct-section-divider">
+  <div class="display-flex flex-justify flex-align-center">
+    <span>
+      <p>
+        TTA Activity:
+        <% tta_reports.each_with_index do |report, i| %>
```
jduss4, line 6: [nit] Not sure you need `each_with_index` since you aren't using `i`, right?

jduss4 (approved):
🥳

app/views/complaints/_tta_report_list.html.erb
Sgtpluck: oh yeah i was gonna do a complicated thing initially with`,` but then laura just told me i could do one per line

PR #149, opened 2021-09-27
Enable using HSES staging data in dev

### LINKED ISSUE #146: Task from #133: Toggle staging data for development
This task is part of #133 

User Story:
As a ~program specialist~ developer, in this case,
I want to see staging data in the complaint tracker
So that I can work with actual information

This task is scoped to:
- [x] use environment variable to use staging data in development
- [x] variable controls all external API connections, not just HSES
- [x] default is to use fake data for development

rahearn:
## Description of change

* Add a single environment variable `CT_USE_REAL_API_DATA` to turn on real API access in development
* Add `dotenv` gem to read configuration from `.env` file
* Add default values to `.env`
* Gem and NPM updates

## Acceptance Criteria

Given the app is running locally with `CT_USE_REAL_API_DATA=true`
When I visit the complaints list
Then I see complaints from HSES's staging environment

## How to test

1. Pull down code
2. Either set `CT_USE_REAL_API_DATA` to true in `.env` or create a new `.env.local` file with the content: `CT_USE_REAL_API_DATA=true`
3. Run `rails server`
4. Visit http://localhost:3000/complaints and verify HSES staging data is listed.

## Issue(s)

* closes #146 

## Checklist

- [x] Meets issue criteria

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


.env
```diff
@@ -0,0 +1,15 @@
...
+export CT_BYPASS_AUTH=false
+
+# CT_CURRENT_USER_UID sets the HSES user id when bypassing authentication
+export CT_CURRENT_USER_UID=fake.testuser@test.com
+
+# CT_CURRENT_USER_NAME sets the user's real name when bypassing authentication
+export CT_CURRENT_USER_NAME="Fake Test-User"
+
+# CT_USE_REAL_API_DATA forces ApiDelegator to connect to real API endpoints
+export CT_USE_REAL_API_DATA=false
```
rahearn, line 15: Definitely looking for feedback about the decision to just put this in `.env` instead of `sample.env`

My thinking was that doing it this way will help with issues I've seen in the past where people add settings to the gitignore'd `.env` and forget to update `sample.env`

.gitignore
```diff
@@ -28,6 +28,9 @@
 # Ignore master key for decrypting credentials and more.
 /config/master.key
 
+# Ignore local configuration overrides
+.env*.local
```
rahearn, line 32: This slightly weird looking format will match both `.env.local` and `.env.<ENV_NAME>.local`

config/environments/development.rb
```diff
@@ -83,5 +83,8 @@
   config.x.hses.auth_base = "https://staging.hses.ohs.acf.hhs.gov"
   config.x.hses.client_id = Rails.application.credentials.hses_client_id
   config.x.hses.client_secret = Rails.application.credentials.hses_client_secret
-  config.x.bypass_auth = ENV["BYPASS_AUTH"] == "true"
+  config.x.bypass_auth = ENV.fetch("CT_BYPASS_AUTH", "false") == "true"
```
rahearn, line 86: This might be overcomplicated now, since we're guaranteeing that `ENV["CT_BYPASS_AUTH"]` will have a value via `dotenv`

.gitignore
Sgtpluck: nice thanks splat

README.md
```diff
@@ -21,6 +21,16 @@ visibility into status and actions taken to address a complaint.
...
 
+#### Local Configuration
+
+Environment variables can be set in development using the [dotenv](https://github.com/bkeepers/dotenv) gem.
+
+Consistent but sensitive credentials should be added to `config/credentials.yml.env` by using `$ rails credentials:edit`
+
+Any changes to variables in `.env` that should not be checked into git should be set
+in `.env.local` and/or `.env.test.local`.
+Both files are required because `.env.local` is loaded for all environments _except_ test.
```
Sgtpluck, line 32: [question] should we be setting local envs for tests from a file, or should we be setting them when they are different within the tests themselves?

README.md
```diff
@@ -32,12 +42,24 @@ HSES Authentication can be bypassed depending on the value of `RAILS_ENV`
...
+
+#### API Connections
+
+API data will come from real APIs or the Api::FakeData depending on the value of `RAILS_ENV`
+
+| Environment | Connects to Real API Endpoints |
+| ----------- | ------------------------------ |
+| test | when `CT_USE_REAL_API_DATA` environment variable is `true` |
+| development | when `CT_USE_REAL_API_DATA` environment variable is `true` |
+| ci | always uses `Api::FakeData` |
```
Sgtpluck, line 60: [nitpick] maybe just `never connects to real API endpoints` since that makes sense with the column header?

config/environments/development.rb
Sgtpluck: [thought] i don't think we need it either way? won't it just return nil if it's not set, and `nil != "true"`

.env
Sgtpluck: [thought] hm, since we're not putting secrets in here or i guess it's fine? in my experience `.env` files are usually hidden since they have secrets and that's the purpose of the sample.env file.

Sgtpluck: [nitpick] technically speaking we don't need to add it here, as false/nil will be handled similiarly, but since we don't have a sample.env it's probably good for awareness that we have it. (especially if we're not documenting ALL env variables in the README)

config/environments/production.rb
```diff
@@ -128,4 +128,7 @@
   config.x.hses.client_id = ENV["HSES_AUTH_CLIENT_ID"]
   config.x.hses.client_secret = ENV["HSES_AUTH_CLIENT_SECRET"]
   config.x.bypass_auth = false
+
+  # API configuration
+  config.x.use_real_api_data = true
```
Sgtpluck, line 133: [thought] probably too late to be thinking this thought, but it might be nice to have these vars/logic so that it doesn't need to do anything special to use real data, so we didn't have to set extra vars in production.
like... `config.x.bypass_real_api` (you can feel free to ignore this thought, as it probably complicates current work, and may not actually simplify anything)

config/environments/test.rb
```diff
@@ -67,5 +67,8 @@
   config.x.hses.auth_base = "https://staging.hses.ohs.acf.hhs.gov"
   config.x.hses.client_id = Rails.application.credentials.hses_client_id
   config.x.hses.client_secret = Rails.application.credentials.hses_client_secret
-  config.x.bypass_auth = ENV["BYPASS_AUTH"] == "true"
+  config.x.bypass_auth = ENV.fetch("CT_BYPASS_AUTH", "false") == "true"
```
Sgtpluck, line 70: [suggestion] as above, `ENV["CT_BYPASS_AUTH"] == "true"` should work as expected

lib/api_delegator.rb
```diff
@@ -15,10 +15,6 @@ def self.namespaced_class(system, endpoint)
   end
 
   def self.namespace
-    ApiDelegator.needs_fake_data? ? Api::FakeData : Api
-  end
-
-  def self.needs_fake_data?
-    Rails.env.development? || Rails.env.test? || Rails.env.ci?
+    Rails.configuration.x.use_real_api_data ? Api : Api::FakeData
```
Sgtpluck, line 18: [praise] i def prefer using a config var here! nice.

README.md
rahearn: Well... certainly within the tests for anything related to the test or that should persist. I was thinking more along the lines of "how do I override something short term to debug" but maybe that's not even worth calling out for `RAILS_ENV=test`.

config/environments/development.rb
rahearn: > i don't think we need it either way?

that's true, it is not strictly needed for the reason you called out. I like having default values (every now and then having a default value instead of nil makes debugging things later easier) but this is now double default value'd

.env
rahearn: Yeah, this file is doing double duty as setting default values and also documenting everything that can be set.

config/environments/production.rb
rahearn: > so that it doesn't need to do anything special to use real data

I'm not sure I'm following. In production I just hardcoded it to always use the real data.

Sgtpluck (approved):

PR #162, opened 2021-10-06
Add timeline class with events

### LINKED ISSUE #96: Link existing TTA Activity Report to Complaint - M
User Story:
As a program specialist
I want to link an existing activity report ID to a complaint
So that I can track progress that was kicked off out-of-band

Acceptance:
When I add an AR ID to the complaints details
Then I can see that AR in the timeline

### Notes

inline form (similar to a search form) for adding the ID, positioned in the summary box on the left side of the complaints details view.

---------------------------
### Error messages:
**If the issue number typed in doesn't match an existing activity report**
This number doesn't match any existing activity reports. Please double-check the number.

**Generic connection issue, and we can't look it up**
We're unable to look up reports in TTA Smart Hub right now. Please try again later.

**Don't have permissions/access to see the activity report**
You do not have permission to access this activity report.

**Activity report isn't for this grantee, assuming we are blocking this behavior**
This activity report is for a different grant recipient. Only activity reports for [Grantee name] may be linked to this complaint.

---------------------------
### Lookup parameters
Should be able to return results on ID or displayID options — needs further definition/discussion @jduss4

Sgtpluck:
## Description of change
Adds the TTA event into the Complaint timeline.
Also adds a timeline class so we can keep track of different types of events (and add more if they come up!)

## Acceptance Criteria
When I add an AR ID to the complaints details
Then I can see that AR in the timeline

## How to test

- [ ] navigate to a complaint
- [ ] add a TTA activity if it does not have one
- [ ] The TTA activity should show up in the timeline in chronological order based on the given date. (Note: because this is currently fake data, if you refresh the page, the date *will* change.)

## Issue(s)

* closes #96 

## Checklist

- [ ] Meets issue criteria

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


app/controllers/complaints_controller.rb
```diff
@@ -13,11 +13,22 @@ def show
...
 
   private
 
+  def tta_reports
+    @issue_tta_reports.pluck(:tta_report_display_id).map do |id|
+      ApiDelegator.use("tta", "activity_report", display_id: id).request[:data]
+    end
+  end
+
+  def tta_report_display_ids
```
rahearn, line 28: [question] where is this method called from?

app/models/api/fake_data/activity_report.rb
```diff
@@ -12,7 +12,8 @@ def data
     @data ||= {
       id: identifier,
       attributes: {
-        display_id: display_id
+        display_id: display_id,
+        startDate: datetime_string
```
rahearn, line 16: [suggestion (nonblocking)] If you're making any other changes anyway, this should be `startDate: date_string`. Otherwise I can make that change when I merge mine in.

app/views/complaints/show.html.erb
```diff
@@ -51,11 +51,11 @@
...
         </h2>
         <ul class="usa-process-list">
-          <% @complaint.events.each do |event| %>
+          <% @timeline.events.each do |event| %>
             <li class="usa-process-list__item usa-process-list__item-disc">
               <h3 class="usa-process-list__heading ct-timeline__two-part-container">
-                <%= @complaint.event_label(event[:name]) %>
-                <span class="ct-timeline__date"><%= event[:date].strftime("%m/%d/%Y") %></span>
+                <%= event.label %>
+                <span class="ct-timeline__date"><%= event.formatted_date %></span>
```
rahearn, line 58: [praise] having a separate class makes these two lines so much nicer to read

spec/models/timeline_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    describe "#init" do
+      it "assigns attr_readers" do
+        event = Timeline::ComplaintEvent.new(event_param)
+        expect(event.name).to eq name
+        expect(event.date).to eq Date.parse(date)
+      end
+    end
+
+    describe "#label" do
+      it "returns the correct label" do
```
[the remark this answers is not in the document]
rahearn: [suggestion] I think testing the logic is way more important than testing all of the key/value pairs. I'd prefer to see one test that grabs the value out of the hash and one test that uses the fallback.

This test is a bit of an issue because it'll pass but we don't know if it's because of the hash or the fallback.

rahearn (changes requested):
Love this structure.

PR #165, opened 2021-10-07
Add error validations when linking TTA Activity Reports

### LINKED ISSUE #139: Error validation for TTA activity report linkage
User Story:
As a program specialist
I want to get appropriate errors when I try to associate a TTA activity report and it fails
So that I can figure out what's going wrong and associate the right reports

Acceptance:
When I try to associate a TTA Activity Report that doesn't exist
Then I get an error that tells me it doesn't exist

This story is associated with
#96 

---------------------------
### Error messages:
**If the issue number typed in doesn't match an existing activity report**
This number doesn't match any existing activity reports. Please double-check the number.

**Generic connection issue, and we can't look it up**
We're unable to look up reports in TTA Smart Hub right now. Please try again later.

**Don't have permissions/access to see the activity report**
You do not have permission to access this activity report.

**Activity report isn't for this grantee, assuming we are blocking this behavior**
This activity report is for a different grant recipient. Only activity reports for [Grantee name] may be linked to this complaint.

rahearn:
## Description of change

* Displays error messages when linking Activity Reports
* Stores HSES access token in session for use in calling TTA Hub API
* Adds test coverage checks
* Adds special cases to Api::FakeData::Tta::ActivityReport to trigger error conditions by using the display id

## Acceptance Criteria

When I try to associate a TTA Activity Report that doesn't exist
Then I get an error that tells me it doesn't exist

## How to test

Attempt to link TTA reports. Ending the display ID with the following numbers will trigger error states when using fake data:
* `-401` or `-500` generic connection issue
* `-403` unauthorized user
* `-404` report doesn't exist
* anything else will return a good activity report

## Issue(s)

* closes #139 

## Checklist

- [x] Meets issue criteria

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


app/controllers/issue_tta_reports_controller.rb
```diff
@@ -1,24 +1,49 @@
...
   def update
-    report = IssueTtaReport.find(params[:id])
-    if report.update(
-      tta_report_display_id: tta_report_display_id,
-      tta_report_id: activity_data[:id]
-    )
-      redirect_to complaint_path(issue_id)
+    if activity_data[:success]
+      activity_report = activity_data[:body][:data]
+      report = IssueTtaReport.find(params[:id])
```
Sgtpluck, line 29: [thought] do we want to guard against an IssueTtaReport not being there?

app/controllers/issue_tta_reports_controller.rb
```diff
@@ -1,24 +1,49 @@
...
   def create
-    @issue_tta_report = IssueTtaReport.new(
-      issue_id: issue_id,
-      tta_report_display_id: tta_report_display_id,
-      tta_report_id: activity_data[:id]
-    )
-
-    if @issue_tta_report.save
-      redirect_to complaint_path(issue_id)
+    if activity_data[:success]
```
Sgtpluck, line 3: [idle thought] this controller is getting so unwieldy that it's starting to feel like maybe we should have an ApiResponse object but also that's probably an overoptimization for right now

app/controllers/issue_tta_reports_controller.rb
```diff
@@ -51,4 +72,15 @@ def tta_report_display_id
   def issue_id
     params[:issue_id]
   end
+
+  def tta_link_error_message(api_response)
+    case api_response[:code]
```
Sgtpluck, line 77: [small suggestion] should we just pass the code through?

app/models/api_request.rb
```diff
@@ -6,10 +6,16 @@ def response
   private
 
   def format_response(res)
+    response_body = begin
+      JSON.parse(res.body)
+    rescue JSON::ParserError
```
Sgtpluck, line 11: [question] is this just a general guard or did you see something specific that would make this happen?

spec/features/complaints/tta_activity_report_spec.rb
```diff
@@ -50,6 +43,38 @@
...
     end
+
+    context "error handling" do
+      before do
+        click_button "Link TTA Activity Report"
+        fill_in "TTA report display ID", with: test_display_id
+        click_button "Link"
+      end
+
+      context "unauthorized user" do
```
Sgtpluck, line 54: [question] i am confused as to how these are working/displaying the correct error. can you please explain? thanks!

spec/requests/complaints_spec.rb
```diff
@@ -10,12 +10,13 @@
...
 
   describe "GET /complaints" do
-    it "returns a 302 status" do
-      get complaints_path
-      expect(response).to have_http_status(302)
-    end
-
     describe "without an authorized user" do
+      around do |example|
+        Rails.configuration.x.bypass_auth = false
```
Sgtpluck, line 15: [praise] nice

app/models/hses_access_token.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_reader :credentials
+
+  def initialize(creds)
+    @credentials = creds.dup.with_indifferent_access
+    @credentials[:access_token] = @credentials.delete :token
+    @credentials.delete :expires
+  end
+
+  def bearer_token
+    access_token.token
```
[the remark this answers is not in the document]
rahearn: exactly 😄

app/models/api/tta.rb
```diff
@@ -1,21 +1,52 @@
...
+  def request
+    Rails.logger.debug <<~EODM
+      TTAHub #{path}:
+      #{JSON.pretty_generate(response)}
+    EODM
+    if response[:code] == "200"
+      response[:body]
+    else
+      Rails.logger.error "TTA Hub call to #{path} responded with #{response[:code]}"
+      fail response[:body].to_json
```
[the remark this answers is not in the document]
rahearn: You are correct that they are aliases.

My personal rule of thumb (that I likely violate in other places) is that `fail` means "This is where a failure originated from" and `raise` means, "I rescued this error and did something with it, but I want to continue it on". Since they're aliases though that distinction is lost on the code and is only for humans

.env
```diff
@@ -35,6 +46,9 @@ export CT_USE_REAL_API_DATA=false
 # HSES_API_PASSWORD is the Basic Auth password to access the HSES API
 # export HSES_API_PASSWORD=
 
+# TTA_HUB_API_HOSTNAME is the host to access the TTA Hub API
+# export TTA_HUB_API_HOSTNAME=
```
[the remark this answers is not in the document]
rahearn: Development defaults to hitting the tta hub staging server, so this only needs to be set to override that.

It's a production ENV var because it'll have to switch from stage to prod

spec/features/complaints/tta_activity_report_spec.rb
rahearn: https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/pull/165/files#diff-7292d49acacbcc2165732b0aad4873886ac24363ff4f152ad9f62f70ac5ee701R13 triggers different error conditions based on the `test_display_id`

rahearn: For each of these tests, the context groups the expectation to see a given error message with the `let(:test_display_id)` that will trigger it, and the `before` block actually does the steps to view the final page.

Sgtpluck: oh hmm i see. i did kind of gloss over that while i was reviewing it. so if the 40X error is in the display ID, it will trigger that error message. would you mind adding a comment to the code about what that's doing, and how it helps with development/testing?

Sgtpluck: maybe both? just to keep future me in the loop

.github/workflows/tests.yml
```diff
@@ -50,6 +50,13 @@ jobs:
           DATABASE_URL: postgres://complaint_tracker:postgres@localhost:5432/complaint_tracker_test
         run: bundle exec rake spec
 
+      - name: Upload coverage information
```
jduss4, line 53: [question] Is the file / artifact being created at `coverage` coming out of the `spec` run step automatically? I'm assuming so, but I want to make sure I understand how that works!

app/models/api/fake_data/tta.rb
```diff
@@ -10,9 +10,46 @@ def initialize(display_id:, access_token:)
     end
 
     def request
-      details_wrapper.merge(
-        data: Api::FakeData::ActivityReport.new(display_id: display_id).data
-      )
+      case display_id
```
jduss4, line 13: [praise] Ooh this is a good idea

README.md
```diff
@@ -49,7 +49,7 @@ HSES Authentication can be bypassed depending on the value of `RAILS_ENV`
 
 | Environment | HSES Bypass |
 | ----------- | ----------- |
-| test | set `CT_BYPASS_AUTH` environment variable to `true` |
+| test | always bypassed |
```
jduss4, line 52: [praise] 👍🏻

app/views/complaints/show.html.erb
```diff
@@ -24,7 +24,7 @@
             </p>
           </div>
         <% end %>
-        <div class="display-none" id="tta-activity-form">
+        <div <%= flash[:tta_errors].blank? && 'class="display-none"'.html_safe %> id="tta-activity-form">
```
jduss4, line 27: [suggestion]
I recommend using `sanitize` rather than `.html_safe`, unless if there's some reason that it won't work here

[question]
I'm not sure I'm following this part. Seems to me like this would print out something like `<div true id="tta-activity-form">` ?  Why do you even need `.html_safe`?  Maybe I'm just unfamiliar with the syntax?

jduss4 (changes requested):
I have a few questions but this looks really nice, and we're definitely going to want to imitate a lot of this implementation for the HSES API, imo!

app/models/api/tta.rb
```diff
@@ -1,21 +1,52 @@
...
+  def initialize(display_id:, access_token:)
+    @display_id = display_id
+    @access_token = access_token
+  end
+
+  def request
+    Rails.logger.debug <<~EODM
+      TTAHub #{path}:
+      #{JSON.pretty_generate(response)}
+    EODM
```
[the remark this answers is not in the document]
rahearn: It means Eagles of Death Metal now, but when I wrote it I chose it for `End Of Debug Message`

.github/workflows/tests.yml
rahearn: Yes, `bundle exec rake spec` runs the coverage stuff automatically

app/views/complaints/show.html.erb
jduss4: For whatever reason, I take the apidock line "It is recommended that you use sanitize instead of this method." very seriously, but I think `.html_safe` works fine here.
https://apidock.com/rails/String/html_safe

I'd prefer something like @Sgtpluck 's suggestion if you don't mind the `class=""` possibility.  I also like `<div class=<%= "display-none" if flash[:tta-errors].blank? %>`.

app/models/issue_tta_report.rb
```diff
@@ -1,5 +1,54 @@
...
+
+  attr_accessor :access_token
+
+  before_validation :retrieve_tta_report_id
+
+  private
+
+  def api_call_succeeded
+    # do not make the api call here, if @activity_data is nil then we didn't attempt to make a call
+    if !@activity_data.nil?
```
Sgtpluck, line 17: [thought] mmmm the conditionals moved but they're still here. but creating a response class it def outside the scope, just a thing to think about for future PRs imo.

however
[suggestion] this could be condensed to
`if activity_data&[:success]` to at least cut down on the double ifs

app/models/issue_tta_report.rb
```diff
@@ -1,5 +1,54 @@
...
+        self.tta_report_id = activity_data[:body][:data][:id]
+      end
+    end
+  end
+
+  def activity_data
+    @activity_data ||= api.request
+  end
+
+  def api
```
Sgtpluck, line 40: [thought] should this be `def get`? it is getting the associated activity report, so i would expect that, i think?

app/models/issue_tta_report.rb
```diff
@@ -1,5 +1,54 @@
...
+    end
+  end
+
+  def errors_blank?
+    errors.blank?
+  end
+
+  def retrieve_tta_report_id
+    if tta_report_display_id_changed?
+      if activity_data[:success]
```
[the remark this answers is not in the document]
rahearn: I think I had a reason for it at the time, but I think a single line makes more sense.

app/models/issue_tta_report.rb
Sgtpluck: tuesday morning me doesn't mind `api` as much as friday afternoon me did. up to you!

app/models/issue_tta_report.rb
```diff
@@ -1,5 +1,50 @@
...
-  validates :issue_id, :tta_report_display_id, :tta_report_id, presence: true
-  validates :tta_report_id, format: {with: /\A[0-9]*\z/, message: "only allows numbers"}
+  validates :issue_id, :tta_report_display_id, presence: true
   validates :tta_report_display_id, uniqueness: {scope: :issue_id}
+  validate :api_call_succeeded
+  validates :tta_report_id, presence: true, if: :errors_blank?
+
+  attr_accessor :access_token
+
+  before_validation :retrieve_tta_report_id
```
Sgtpluck, line 11: so just to be make sure i understand what's happening here ...
`retrieve_tta_report_id` happens before `.save` runs. then `api_call_succeeded` runs. if there are no errors, than it runs through the validators?

is that the correct passage? this is hard to get at a glance -- can we maybe lump those lines together or in some other way make this a bit more clear?

app/views/complaints/_tta_create_activity_report.html.erb
```diff
@@ -1,15 +1,24 @@
 <div class="ct-section-divider">
```
Sgtpluck, line 1: [question] do we need a similar thing for the `_tta_edit_activity_report` partial? (also did i goof -- should that be `update` and not `edit`?)

spec/models/issue_tta_report_spec.rb
```diff
@@ -5,8 +5,7 @@
     subject do
       described_class.new(
         issue_id: "ISSUE_ID",
-        tta_report_display_id: "ACTIVITY_REPORT_ID",
-        tta_report_id: "12345"
```
Sgtpluck, line 9: [suggestion] i think we need another comment here about how it is retrieving the tta_report_id based on the display_id regex match, as i again didn't get it until i looked more closely at the test data.

Sgtpluck (changes requested):
i think this is very close! just a few more suggestions to make things clearer; also, can you please update the ffaker_wrapper tests with the new ffaker methods you added?? thank you!

app/views/complaints/_tta_create_activity_report.html.erb
rahearn: So... I think we need to break this off into another story. Right now errors on edit will open this form instead of but I think fixing that will require a not insignificant number of changes.

As for the naming, if anything I think this one should be `_tta_new_activity_report.html.erb` rather than changing the edit one, but both forms work for me

Sgtpluck: works for me!

Issue #134, opened 2021-09-16
Task from #133: Complaints pagination backend


jduss4:
This task is part of https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/133 : incorporating HSES staging data in the complaint tracker

This task is scoped to:
- [ ] Adding limit / offsets to the API request
- [ ] Frontend pagination (could split into separate task)

lauranash18f:
Using USWDS component — may want to add number of results in future
![Complaints Tracker: Pagination and Filters - All Region 1](https://user-images.githubusercontent.com/85581840/135127394-31468f1d-e44a-48e3-a2e2-9a00c769a172.png)
.

PR #170, opened 2021-10-08
Pagination backend

jduss4:
## Description of change

Adds methods to calculate offset and limit from page request to complaints list
Creates default parameters in ApiRequest so that subclasses (like Hses::Issues) need not worry about them,
and instead use `parameters` method to provide their own options

Note: the functionality of `query` was moved to `parameters` and `query` now combines default parameters and subclass parameters

Another note: in doing this, I realized that when HSES changed the API spec to use `form` and removed `deepObject`, we seem to have lost the URL params we were hoping for, namely: we wanted `page[offset]=x&page[limit]=x` but the current API wants `offset=x&limit=x`. This is something we may want to revisit.

## Acceptance Criteria

- Passes limit / offset information to the HSES API

## How to test

~Well, manually you can drop `page=2` or whatever you fancy into the `/complaints` URL and see the results change.~ (removed after comment)

## Issue(s)

* closes https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/134

## Checklist

- [X] Meets issue criteria
- [?] Code is meaningfully tested
- ~[] Meets accessibility standards (WCAG 2.1 Levels A, AA)~
- ~[] API Documentation updated~

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


app/models/api_request.rb
```diff
@@ -21,14 +31,30 @@ def host
     raise "Please define a host method in #{self.class}"
   end
 
+  def page_limit
```
Sgtpluck, line 34: [suggestion] 
having this conditional logic feels like a smell to me. it would be more OO to pull the logic into Api::Hses (or Api::Hses::Issues, whichever feels more appropriate) and have these methods return `""` so that they will be more configurable in different endpoints/systems.

other options:
* default_params could be just ["username=#{@username}&types=1"] and more default_params could be defined within the individual systems.
* if we are using page_limit and page_offset for all systems define it further upstream so we don't have to use the conditional

app/controllers/complaints_controller.rb
```diff
@@ -4,7 +4,7 @@ class ComplaintsController < ApplicationController
   before_action :check_pa11y_id, only: :show, if: -> { Rails.env.ci? }
 
   def index
-    api = ApiDelegator.use("hses", "issues", {user: session["user"]})
+    api = ApiDelegator.use("hses", "issues", {user: session["user"], page: params[:page]})
```
Sgtpluck, line 7: [thought] this seems unnecessary for the scope of this PR, since there is a default keyword and we don't have it working yet, and feels like it could make limit whoever does the frontend. (although that'll probably be you so maybe it's fine) am i missing how this work?

spec/models/api_request_spec.rb
```diff
@@ -28,25 +28,64 @@
...
+      before do
+        success = Net::HTTPSuccess.new(1.0, "200", "OK")
+        allow_any_instance_of(Net::HTTP)
+          .to receive(:start)
+          .and_return(success)
+        expect(success)
+          .to receive(:body)
+          .and_return('{"meta":{},"data":[]}')
+
+        # mock in host, path, and query
```
Sgtpluck, line 64: [nitpick] `parameters`

spec/models/api_request_spec.rb
```diff
@@ -28,25 +28,64 @@
...
+        allow_any_instance_of(ApiRequest).to receive(:host).and_return "example.com"
+        allow_any_instance_of(ApiRequest).to receive(:path).and_return "/path"
+        allow_any_instance_of(ApiRequest).to receive(:parameters).and_return []
+      end
+
+      it "returns no offset if no page is set" do
+        api.response
+        uri = api.instance_variable_get(:@uri)
```
Sgtpluck, line 72: [thought] hmm this feels like we are testing private variables, which we shouldn't be doing.

instead we should be checking to make sure that `URI::HTTPS.build` and  `Net::HTTP.start` are being called with the correct arguments.

`expect(URI::HTTPS).to receive(:build).with("example.com", "/path", "username=#{@username}").and_return $some_mocked_object`

Sgtpluck: (this is pseduo code, happy to work with you on it if this doesn't work quite right, i always have to look up how to do this)

spec/models/api_request_spec.rb
```diff
@@ -28,25 +28,64 @@
...
+        expect(uri.to_s).not_to include("offset")
+        expect(uri.to_s).not_to include("limit")
+      end
+      it "returns the correct offset and limit if page is set to 1" do
+        api.instance_variable_set(:@page, 1)
```
Sgtpluck, line 77: [thought] having to set this in this way makes me feel even more sure that this should be set in the system Apis rather than in the general ApiRequest class!

Sgtpluck (changes requested):

spec/models/api_request_spec.rb
jduss4: I tried that, because that was your suggestion the other day when I asked, and ended up banging my head against an annoying amount of things before I gave up and went with `@uri` instead.  I don't feel great about it, though.  I think if we need this PR in right away, then someone else may have to pick up fixing these tests because I can't spend a lot of time on it today, but otherwise if next week would work I would definitely revisit it with you!

spec/models/api_request_spec.rb
jduss4: thanks!

app/models/api_request.rb
```diff
@@ -1,10 +1,20 @@
 class ApiRequest
+  PAGE_LIMIT = 25
+
   def response
     @response ||= get_response
   end
 
   private
 
+  def default_params
```
rahearn, line 10: [suggestion] I think it will be unlikely that all the systems will agree enough on the params for this to be useful. At best, it's going to result in unnecessary parameters being sent. At worst it'll result in weird bugs. Even within a system you won't need paging params for show routes

app/models/api_request.rb
```diff
@@ -21,14 +31,30 @@ def host
...
 
   # Inheriting class defines
+  def parameters
+    raise "Please define a parameters method in #{self.class}"
+  end
+
   def query
-    raise "Please define a query method in #{self.class}"
+    req_params = default_params + parameters
+    req_params.compact.join("&")
```
rahearn, line 57: [thought] can the query params array be passed directly into either the `URI` classes or `Net::HTTP.start` call? Then those classes could deal with CGI escapes properly. If it can't then:

[suggestion] I am a fan of a single (overridable) method that takes the parameters array and turns it into a useable query string, but we should use the `CGI` class(? module?) to properly escape and join them.

spec/models/api_request_spec.rb
rahearn: [suggestion] I agree with testing via expectations like Davida said. If that won't work, for whatever reason, then I'd rather pull the `request_uri` method to public and test that rather than reaching in to the instance variable.

jduss4: I think I've done it!  By jove.  I figured out what I was doing wrong, too, so that's great news for my future

spec/models/api/hses_spec.rb
```diff
@@ -21,28 +21,86 @@
   end
 
   describe "Issues" do
-    let(:issues) { Api::Hses::Issues.new(user: user) }
+    context "with params argument" do
+      let(:issues) {
+        Api::Hses::Issues.new(
+          user: user,
+          params: ActionController::Parameters.new({page: "2"})
```
Sgtpluck, line 28: i don't think i've used `ActionController::Parameters` directly before. is there an advantage to using that, or just passing in the hash directly?

app/models/api/tta.rb
```diff
@@ -44,7 +44,7 @@ def path
     "/api/v1/activity-reports/display/#{display_id}"
   end
 
-  def query
-    ""
+  def parameters
+    []
```
rahearn, line 48: [issue(?)] should this be an empty hash? The comment in `ApiRequest` indicates that it should be a hash

spec/models/api_request_spec.rb
```diff
@@ -3,45 +3,64 @@
 
 RSpec.describe ApiRequest do
   describe "#response" do
+    it "is memoized" do
+      api = ApiRequest.new
+      expect(api).to receive(:get_response).once.and_return "value"
+      expect(api.response).to eq api.response
+    end
+
     describe "with valid host, path, and query" do
```
rahearn, line 12: [nitpick] 

```suggestion
    describe "with valid host, path, and parameters" do
```

spec/models/api_request_spec.rb
```diff
@@ -3,45 +3,64 @@
...
-        allow_any_instance_of(ApiRequest).to receive(:query).and_return "?some=query"
+        # mock in host, path, and configure auth
+        allow(@api).to receive(:host).and_return "example.com"
+        allow(@api).to receive(:path).and_return "/path"
+        allow(@api).to receive(:configure_auth)
+      end
 
-        res = ApiRequest.new.response
+      it "returns an object with a code and a body" do
+        allow(@api).to receive(:parameters).and_return []
```
rahearn, line 29: [suggestion] another non-hash. Either the comment or code should be updated

spec/models/api_request_spec.rb
```diff
@@ -3,45 +3,64 @@
...
+          .to receive(:parameters)
+          .and_return({one: 1, two: ["a", "b"]})
+        expect(URI::HTTPS)
+          .to receive(:build)
+          .with({
+            host: "example.com",
+            path: "/path",
+            port: 443,
+            query: "one=1&two=a&two=b"
+          }).and_call_original
```
rahearn, line 45: [praise] _love_ this use of expectations and also using the real implementation

spec/models/api/hses_spec.rb
```diff
@@ -21,28 +21,86 @@
...
+        )
+      }
 
-    describe "#initialize" do
-      it "sets @username" do
-        expect(issues.instance_variable_get(:@username)).to eq user["uid"]
+      describe "#initialize" do
+        it "sets @username" do
+          expect(issues.instance_variable_get(:@username)).to eq user["uid"]
```
Sgtpluck, line 34: [suggestion] mmm, if it's not publicly available, we shouldn't be reaching in to grab it. i don't think we need these tests!

spec/models/api/hses_spec.rb
```diff
@@ -21,28 +21,86 @@
...
+      describe "#request" do
+        it "returns a Ruby object with meta and data keys" do
+          # stub the net/http request / response
+          response = Net::HTTPSuccess.new(1.0, "200", "OK")
+          expect_any_instance_of(Net::HTTP).to receive(:start).and_return response
+          expect(response).to receive(:body).and_return('{"meta":{},"data":[]}')
+
+          expect(issues.request).to match({"meta" => {}, "data" => []})
+        end
+        it "sends query parameters for type, username, and pagination" do
```
Sgtpluck, line 88: [annoying nitpick] i prefer a newline in between blocks to make it easier to read. thanks!

spec/models/api_request_spec.rb
Sgtpluck: ➕

spec/models/api_request_spec.rb
Sgtpluck: ➕

spec/models/api_request_spec.rb
```diff
@@ -3,45 +3,64 @@
...
-    describe "without combinations of configure_auth, host, path, and query" do
+    describe "without combinations of configure_auth, host, path, and parameters" do
       it "returns an error for missing host" do
         allow_any_instance_of(ApiRequest).to receive(:path).and_return "/path"
-        allow_any_instance_of(ApiRequest).to receive(:query).and_return "?some=query"
+        allow_any_instance_of(ApiRequest).to receive(:parameters).and_return []
         allow_any_instance_of(ApiRequest).to receive(:configure_auth)
         expect { ApiRequest.new.response }.to raise_error(RuntimeError)
       end
       it "returns an error for missing path" do
```
Sgtpluck, line 57: [nitpick] same nit about newlines in between these tests.

app/models/api/tta.rb
jduss4: Oh snap, good catch yes, it's a hash -- I should change that and the tests stubbing that in (even though it's not used in them)

spec/models/api_request_spec.rb
jduss4: Yeah, apparently I was thinking back to the glory days of default_parameters + parameters or something, my bad

spec/models/api/hses_spec.rb
jduss4: I was just trying to be as realistic as possible, since presumably if we're passing parameters in from the controller that's the type we would be getting. The test will be fine with just a hash if you have a preference for that.

PR #173, opened 2021-10-12
Grantee details skeleton

### LINKED ISSUE #168: Grantee details skeleton - S
User Story:
As a program specialist
I want to navigate to the details page for a grantee
So that I have one jumping off place to get more information

Acceptance:
Given that I am a Program Specialist
When I click the grantee's name from the complaints table or details
Then I see a page skeleton for the grantee

Prerequisite for #119 

Scoped to:

* basic USWDS components with only content that we already have in the system and placeholders.
* only "outer" box for complaints list. Inner details and design will take place when adding the complaints

kbighorse:
## Description of change
- unit tests for grantee model
- create grantee model
- create route to `/grantee/:id`
- create details page for a single grantee
- note: `/grantees` is out of scope
- integration test of required elements

## Acceptance Criteria
- visit `/complaints`
- visit `/grantees/:id` link in complaint in `Grantee` column
- observe details header
- observe left sidebar:
  - Summary card
  - Open issues card
- observe skeletal list of complaints

## How to test
`rake spec`

## Issue(s)

* closes #168 

## Checklist

- [ ] Meets issue criteria
- [ ] Boundary diagram updated

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/spec/models/complaint_spec.rb
+++ b/spec/models/complaint_spec.rb
@@ -30,8 +30,9 @@ RSpec.describe Complaint do
     end
   end
 
+  # TODO: test for missing attribute
   describe "#grantee" do
-    it "deletages to the attributes" do
+    it "delegates to the attributes" do
       expect(subject.grantee).to eq hses_complaint[:attributes][:grantee]
     end
   end
@@ -145,6 +146,7 @@ RSpec.describe Complaint do
     end
   end
 
+  # TODO: test for missing attribute
   describe "#summary" do
     it "delegates to the attributes" do
       expect(subject.summary).to eq hses_complaint[:attributes][:summary]


config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  resources :grantees
```
Sgtpluck, line 2: [suggestion] only define the route we are currently using, so
`resources :grantees, only: %i[show]`

app/helpers/grantees_helper.rb
```diff
@@ -0,0 +1,5 @@
+module GranteesHelper
```
Sgtpluck, line 1: [suggestion] although we discussed not doing FakeData in this change, since it's done and you included a Grantee model, we don't need this helper, the grantee name is accessible directly from the model

app/models/grantee.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def name
+    attributes["name"]
+  end
+
+  def region
+    attributes["region"]
+  end
+
+  def name?
```
Sgtpluck, line 22: [suggestion] looks like this method is currently not invoked, so can be removed!

spec/features/grantees/grantee_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
```
Sgtpluck, line 1: [suggestion] please remove this file, as there are no tests in it

spec/helpers/grantees_helper_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
```
Sgtpluck, line 1: [suggestion] please remove this file when you remove the grantee_helper!

spec/models/grantee_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe Grantee, type: :model do
+  let(:hses_grantee) { Api::FakeData::Grantee.new.data }
+  subject { described_class.new hses_grantee }
+
+  before do
+    hses_grantee[:attributes].merge!({
```
Sgtpluck, line 8: [question] what is this `before` block doing? can we get rid of it?

spec/models/grantee_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    it "sets @id" do
+      expect(subject.id).to eq hses_grantee[:id]
+    end
+
+    it "sets @attributes" do
+      expect(subject.attributes).to eq hses_grantee[:attributes].with_indifferent_access
+    end
+  end
+
+  describe "#name" do
```
Sgtpluck, line 22: [thought] hmmm in these tests (and the `complaint_spec` which i assume these are based off) we don't really test the unhappy path. we probably should think about adding tests that indicate what we want to happen if the attribute is missing.

spec/requests/grantees_spec.rb
```diff
@@ -0,0 +1,130 @@
+require "rails_helper"
+
+# This spec was generated by rspec-rails when you ran the scaffold generator.
```
Sgtpluck, line 3: [suggestion] please delete all commented code, and code related to a route that is not the show route

spec/requests/grantees_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+# controller code, this generated spec may or may not pass.
+#
+# It only uses APIs available in rails and/or rspec-rails. There are a number
+# of tools you can use to make these specs even more expressive, but we're
+# sticking to rails and rspec-rails APIs to keep things simple and stable.
+
+RSpec.describe "/grantees", type: :request do
+  # Grantee. As you add validations to Grantee, be sure to
+  # adjust the attributes here as well.
+  let(:valid_attributes) {
```
Sgtpluck, line 18: [suggestion] either delete these or use them, please!

spec/requests/grantees_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+  # describe "GET /index" do
+  #   it "renders a successful response" do
+  #     Grantee.create! valid_attributes
+  #     get grantees_url
+  #     expect(response).to be_successful
+  #   end
+  # end
+
+  describe "GET /show" do
+    let(:grant_id) { FakeIssues.instance.data.first[:id] }
```
Sgtpluck, line 35: [suggestion] i think this should be `FakeIssues.instance.data.first[:grantNumber]` to make this a little more clear

spec/routing/grantees_routing_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  describe "routing" do
+    it "routes to #index" do
+      expect(get: "/grantees").to route_to("grantees#index")
+    end
+
+    it "routes to #new" do
+      expect(get: "/grantees/new").to route_to("grantees#new")
+    end
+
+    it "routes to #show" do
```
Sgtpluck, line 13: [suggestion] please get rid of all the tests unrelated to the `show` route, thanks!

app/controllers/grantees_controller.rb
```diff
@@ -0,0 +1,73 @@
+class GranteesController < ApplicationController
+  before_action :set_grantee, only: %i[show edit update destroy]
```
Sgtpluck, line 2: [suggestion] since we only have one route/method here currently, i'd prefer this not be abstracted out, and instead be set within the `show` method.

it also would be good to have the ID that is passed through the params be the grantee_id, rather than having it be random

app/models/api/fake_data/grantee.rb
```diff
@@ -0,0 +1,20 @@
+require "ffaker_wrapper"
+
+class Api::FakeData::Grantee
```
Sgtpluck, line 3: [suggestion] we probably want to do something similar to the ActivityReport where we pass in the ID to an initializer so it can be consistent with the grantee params

app/views/grantees/show.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  <div class="grid-row grid-gap">
+    <div class="grid-col-4">
+      <div class="summary-section">
+        <div class="ct-section-divider">
+          <h2>Summary</h2>
+        </div>
+      </div>
+    </div>
+    <div class="grid-col-8">
+      <ul class="usa-process-list">
```
Sgtpluck, line 17: [suggestion] it looks like this is just a list of complaints, rather than the complaint containers. (if this feels like it's gotten too big, though, this can be done in the next card, though)

app/helpers/grantees_helper.rb
kbighorse: 👍🏾

app/models/grantee.rb
kbighorse: 👍🏾

spec/requests/grantees_spec.rb
kbighorse: This is actually wrong, it's not the `id` at all! Thanks for the catch.

Sgtpluck: The FakeIssues is a list of fake data coming from the `/complaints` endpoint, so the top-level ID is the ID of an issue, not a grant or a grantee.

If we want the grant id out of that object, it looks like that should be `first[:attributes][:grantNumber]`. (That being said, I don't think we need to use the FakeIssues to get a grant id, and could probably just use a random identifier, although you'd need to test to make sure)

kbighorse: oh ok, is there a `/grantees` endpoint yet? I suppose a `FakeGrantees` source should be created, so we're not going through `FakeIssues`, probably in its own ticket.

app/views/grantees/show.html.erb
kbighorse: @lauranash18f is going to give me some guidance on the component layout

app/views/grantees/show.html.erb
```diff
@@ -0,0 +1,39 @@
...
+</div>
+<div class="details-body grid-container">
+  <div class="grid-row grid-gap">
+    <div class="grid-col-4">
+      <div class="summary-section">
+        <div class="ct-section-divider">
+          <h2>Summary</h2>
+        </div>
+      </div>
+      <div class="grid-col-4">
```
Sgtpluck, line 15: [suggestion] this div is unnecessary -- the Open Issues container is part of the `grid-col-4` div. by adding the other column, it is trying to divide itself smaller within the container grid i think.

if you remove this `div`, the Open Issues container should fill the whole column like the `summary` container.

spec/models/grantee_spec.rb
kbighorse: adding TODO comments

PR #178, opened 2021-10-14
Add ApiResponse object

Sgtpluck:
## Description of change
This is a spike on refactoring the API response to add an ApiResponse class. Thoughts?

## Issue(s)

* closes #0

## Checklist

decisions

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/models/api_response.rb
+++ b/app/models/api_response.rb
@@ -14,7 +14,7 @@ class ApiResponse
   end
 
   def body
-    failed? ? {} : JSON.parse(@response_body).with_indifferent_access
+    succeeded? ? JSON.parse(@response_body).with_indifferent_access : {}
   end
 
   def data
--- a/spec/models/api/tta_spec.rb
+++ b/spec/models/api/tta_spec.rb
@@ -47,7 +47,7 @@ RSpec.describe Api::Tta do
           expect(subject.request.body).to eq({})
         end
 
-        it "has a code of 500" do
+        it "has a code of 404" do
           expect(subject.request.code).to eq 404
         end
 
--- a/spec/models/api_response_spec.rb
+++ b/spec/models/api_response_spec.rb
@@ -3,7 +3,9 @@ require "fake_issues"
 
 RSpec.describe ApiResponse do
   context "a successful response" do
-    let(:params) { Net::HTTPSuccess.new(1.1, "200", "OK") }
+    let(:http_obj) { Net::HTTPSuccess.new(1.1, "200", "OK") }
+    subject { ApiResponse.new(http_obj) }
+
     let(:response_body) do
       {
         data: {
@@ -14,10 +16,9 @@ RSpec.describe ApiResponse do
         }
       }
     end
-    let(:subject) { ApiResponse.new(params) }
 
     before do
-      expect(params).to receive(:body).and_return(JSON.generate(response_body))
+      expect(http_obj).to receive(:body).and_return(JSON.generate(response_body))
     end
 
     describe "#initialize" do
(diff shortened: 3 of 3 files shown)


jduss4:
I love it!!!  I think this is a great proof of concept and would love to start adding a few other methods to it as we go, like @rahearn suggested above.  Thanks for working on this!  🥳

spec/models/api/tta_spec.rb
```diff
@@ -9,19 +9,29 @@
...
+            grantee: "Some Grantee"
+          }}
+        end
+
         before do
           # stub the net/http request / response
           response = Net::HTTPSuccess.new(1.1, "200", "OK")
           expect_any_instance_of(Net::HTTP).to receive(:start).and_return response
-          expect(response).to receive(:body).and_return('{"data":{}}')
+          expect(response).to receive(:body).and_return(JSON.generate(response_data))
```
rahearn, line 22: [praise] love pulling this out into `let(:response_data)`

spec/models/api_request_spec.rb
```diff
@@ -20,19 +21,23 @@
         expect(success).to receive(:body).and_return('{"meta":{},"data":[]}')
 
         # mock in host, path, and configure auth
-        allow(@api).to receive(:host).and_return "example.com"
-        allow(@api).to receive(:path).and_return "/path"
-        allow(@api).to receive(:configure_auth)
+        allow(api).to receive(:host).and_return "example.com"
```
rahearn, line 24: [praise] thank you for moving these out of instance variables

spec/models/api_response_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+      {
+        data: {
+          grantee: "Some Grantee"
+        },
+        meta: {
+          itemTotalCount: 20
+        }
+      }
+    end
+    let(:subject) { ApiResponse.new(params) }
```
jduss4, line 17: [suggestion] I wonder if you would consider renaming `params` here? I was confused about why params would be getting passed to the ApiResponse because I was conflating the term with controller params. I think I would be less confused if this was `response` or `res` or `http_obj` or many other terms!

rahearn, line 17: [suggestion (nonblocking)] the `subject` is always setup the same, so I'd prefer pulling it to the top of the describe block.

[nitpick (nonblocking)] Pretty sure `subject` used here is an alias to `let(:subject)` but I like setting the subject with just `subject { ApiResponse.new(params) }`

app/models/api_response.rb
```diff
@@ -0,0 +1,23 @@
...
+  def succeeded?
+    code == 200
+  end
+
+  def failed?
+    !succeeded?
+  end
+
+  def body
+    failed? ? {} : JSON.parse(@response_body).with_indifferent_access
```
jduss4, line 17: [suggestion] Maybe consider reversing this to be `succeeded? ? JSON.parse(...) : {}` ?

spec/models/api_request_spec.rb
```diff
@@ -10,8 +10,9 @@
     end
 
     describe "with valid host, path, and configure_auth" do
+      let(:api) { ApiRequest.new }
```
jduss4, line 13: [huh, non-blocking] I thought I tried this when I wrote this test, and finally decided to stick it in as a `@` instead when it wouldn't work. Maybe I was trying to use `let` in the before block and am misremembering!  HMMM.  Well, good to know it works.

jduss4 (approved):

spec/models/api/tta_spec.rb
```diff
@@ -33,16 +43,16 @@
...
-              title: "Not Found",
-              detail: ""
-            }
-          })
+        it "returns an empty hash as the body" do
+          expect(subject.request.body).to eq({})
+        end
+
+        it "has a code of 500" do
+          expect(subject.request.code).to eq 404
```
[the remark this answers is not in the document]
kbighorse: 👍🏾 resolve at your convenience

PR #181, opened 2021-10-14
Use ApiResponse class in FakeData responses

### LINKED ISSUE #180: ApiResponse not used in FakeData
FakeData responses and real API responses have fallen out of sync with each other.

rahearn:
## Description of change

Use ApiResponse in FakeData and update controllers and models that interact with responses.

## Issue(s)

* closes #180

## Checklist

- [x] Meets issue criteria

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/models/api/hses.rb
+++ b/app/models/api/hses.rb
@@ -42,7 +42,7 @@ class Api::Hses::Issues < ApiRequest
   end
 
   def request
-    response.body
+    response
   end
 
   private
--- a/spec/models/api/hses_spec.rb
+++ b/spec/models/api/hses_spec.rb
@@ -65,7 +65,7 @@ RSpec.describe Api::Hses do
           expect_any_instance_of(Net::HTTP).to receive(:start).and_return response
           expect(response).to receive(:body).and_return('{"meta":{},"data":[]}')
 
-          expect(issues.request).to match({"meta" => {}, "data" => []})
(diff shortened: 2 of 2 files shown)


app/models/issue_tta_report.rb
```diff
@@ -15,8 +15,8 @@ class IssueTtaReport < ApplicationRecord
 
   def api_call_succeeded
     # do not make the api call here, if @activity_data is nil then we didn't attempt to make a call
-    if !@activity_data.nil? && !@activity_data[:success]
-      errors.add(:base, tta_link_error_message(@activity_data[:code]))
+    if !@activity_data.nil? && @activity_data.failed?
```
Sgtpluck, line 18: [question] in what cases would @activity_data be nil? it looks like if it doesn't exist than we hit the endpoint.

rahearn: Any cases where we called `.valid?` or `.save` without updating the display id. I don't believe that's an actual code-path currently.

lib/fake_api_response_wrapper.rb
```diff
@@ -1,4 +1,12 @@
+require "ostruct"
```
jduss4, line 1: [praise] TIL!  I have never had occasion to use this before but it's pretty cool

Issue #205, opened 2021-10-26
Task for 201: Associate monitoring review with complaint


jduss4:
This task is part of #201 

Full User Story:
As a program specialist
I want to manually link a complaint to a monitoring review (in IT-AMS)
So that I can see basic information about the review when I view a complaint

This task is scoped to:
- [ ] imitating TTA activity report UI and behavior for adding a monitoring review to a specific complaint

Issue #207, opened 2021-10-26
Task from 201: Mock errors for monitoring review association with complaint


jduss4:
This task is part of #201 

Full User Story:
As a program specialist
I want to manually link a complaint to a monitoring review (in IT-AMS)
So that I can see basic information about the review when I view a complaint

This task is scoped to:
- [ ] imitating TTA activity report fake data behavior for testing errors in response from monitoring

jduss4:
See this convo for some changes we should probably make:

https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/pull/214#discussion_r739445810

jduss4:
There will not be an integration set up with IT-AMS for this demo product, so this issue is moving back to the backlog for the time being. We can set it up to work with fake data but it's not a priority.

jduss4:
Will need to address the TODOs in `issue_monitoring_reviews_controller.rb` for create_errors and update_errors

PR #214, opened 2021-10-29
Add fake monitoring link to complaint

jduss4:
## Description of change

Copy pasted tta_activity_report into parallel monitoring_review files and then removed or ignored editing / deleting / error handling behavior. Would have liked to generalize more shared partials, but the TTA display_id behavior was just different enough I decided to leave it alone for now.

Also changed the faux-active model validation error handling in favor of a custom `valid?` method for the sake of making things make more sense to me, although I very much expect I have missed a good reason why TTA was set up that way.

Reorganized the UI a little to accommodate TTA + monitoring in the sidebar, slight refactor to JS.

## Acceptance Criteria

As a program specialist
I want to manually link a complaint to a monitoring review (in IT-AMS)
So that I can see basic information about the review when I view a complaint

## How to test

Start app, go to complaint detail page, add a new RAN id.

## Issue(s)

* closes https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/205

## Checklist

- [X] Meets issue criteria
- [X] Code is meaningfully tested (could probably use more)

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**


app/controllers/issue_monitoring_reviews_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      issue_id: issue_id,
+      review_id: review_id,
+      access_token: hses_access_token
+    )
+
+    respond_to do |format|
+      if @issue_monitoring_review.save
+        format.js { render inline: "location.reload(true);" }
+      else
+        Rails.logger.error("Problem saving the link to RAN review: " + @issue_monitoring_review.errors.full_messages.join("; "))
```
jduss4, line 15: [note] I can fill this sort of section out in this PR but I was hoping to save it for https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/207

app/javascript/packs/complaint-links.js
```diff
@@ -14,22 +19,43 @@ const closeForm = function (containerId = "#tta-activity-form") {
...
-  openForm()
+  openForm("#tta-activity-form", "#tta-report-display-id")
 })
 
 document.querySelector("#tta-report-icon").addEventListener('click', event => {
-  openForm()
+  openForm("#tta-activity-form", "#tta-report-display-id")
 })
 
+// ITAMS open form
```
jduss4, line 33: [question] I decided it was probably more readable just to leave the TTA and monitoring blocks as totally separate event listeners, rather than try to create something like `documentQuerySelectorAll("#tta-report-icon, #monitoring-review-icon")...` type code.  What do you think?

app/models/monitoring_review.rb
```diff
@@ -0,0 +1,33 @@
...
+    @id = id
+    @access_token = access_token
+    @errors = []
+  end
+
+  def start_date
+    get_api_data_field :attributes, :reviewStartDate
+  end
+
+  def valid?
```
jduss4, line 16: [question] This is the part where I veered off from the active model validations and went offroading.  I'm not sure it's going to be successful in the long-run but I found it much more easy to understand what was happening. What do folks think?  Should I stick to how TTA is doing it (and therefore possibly pull some of the shared pieces into a concern) or will this be an okay way of doing things?

app/models/monitoring_review.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  private
+
+  def api
+    ApiDelegator.use("monitoring", "review", id: id, access_token: access_token)
+  end
+
+  def get_api_data_field(*path)
+    review_data.data.dig(*path)
```
jduss4, line 27: [question] I didn't really follow the logic of adding errors per field lookup, rather than for the actual API response not being valid, so I simplified this method.  Please let me know if I'm missing something important about the field by field error adding approach!

app/views/complaints/_complaints_open_new_linked_item.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="display-flex flex-align-center">
```
jduss4, line 1: [question] This is the kind of approach that I could take if we want me to keep combining partials. Since this was such a small chunk it seemed like a good test case, and it's still fairly readable.  Not sure that some of the bigger partials are going to be nearly as readable?  But I am happy to give it a try if y'all want.

app/controllers/issue_monitoring_reviews_controller.rb
rahearn: [suggestion] maybe a middle ground of `format.js { render inline: "alert('there was an error');" }`

app/controllers/issue_tta_reports_controller.rb
```diff
@@ -42,6 +42,6 @@ def tta_report_display_id
   end
 
   def issue_id
-    params[:issue_id]
+    params[:issue_id_tta]
```
rahearn, line 45: [question] why did this need to change?

app/javascript/packs/complaint-links.js
rahearn: Yeah, I think the value of clarity in the `openForm` arguments makes it worthwhile to split up

app/models/monitoring_review.rb
rahearn: I think this is fine for now. Once more error reporting is added (and the real API is nailed down) we might be able to converge a bit more. I think doing that now is premature.

app/models/monitoring_review.rb
rahearn: The field-by-field was really needed when displaying the timeline that had an "invalid for this user" report to be able to still show that report but not link out to it or display the topics and creators.

Since errors are explicitly out of scope for this PR I think this is fine for now and will likely need to change later.

app/views/complaints/_complaints_open_new_linked_item.html.erb
rahearn: I think it could make sense to do that for the _issue_tta_report and _issue_monitoring_review at least, but maybe that can/should also be punted until editing of the reviews is enabled

spec/models/issue_monitoring_review_spec.rb
```diff
@@ -26,6 +26,7 @@
 
     it "is not valid without a start date" do
       subject.start_date = nil
+      allow(subject.monitoring_review).to receive(:valid?).and_return(false)
```
rahearn, line 29: [praise] I think this is a good way to go about this, and will work with both your method and `ActiveModel::Validations`

app/controllers/issue_monitoring_reviews_controller.rb
jduss4: Good idea!

app/controllers/issue_tta_reports_controller.rb
jduss4: When I started working on this, there was a hidden input with id="issue_id" for the TTA form. I copied the whole section and started working on monitoring and totally didn't think about changing that ID.  Then I got this:

https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/pull/214#issuecomment-954893194

No duplicate ids -- so I figured I would rather change them both to be `issue_id_{system}` rather than leave TTA as is.  I'm open to other ideas, though, if you would prefer something else!

app/models/monitoring_review.rb
jduss4: Oooo I forgot about the timeline aspect. I will add a comment referencing this in the error issue.

app/views/complaints/_complaints_open_new_linked_item.html.erb
jduss4: There is definitely a lot in terms of editing / deleting that will need to be in the shared partials, so punting makes sense to me!

app/controllers/issue_tta_reports_controller.rb
rahearn: Ah, that makes sense. I'm good with doing things this way

rahearn (approved):
🎉

PR #215, opened 2021-10-29
Wire up complaint endpoint

Sgtpluck:
## Description of change
This wires in the single detail issue endpoint

## Acceptance Criteria
The complaint detail still works!

## How to test

- [ ] set local env to `export CT_USE_REAL_API_DATA=true
- [ ] restart your local server
- [ ] navigate to a complaint
- [ ] real data should be getting pulled in

## Issue(s)

* closes #0

## Checklist

decisions

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/models/api/hses.rb
+++ b/app/models/api/hses.rb
@@ -16,21 +16,35 @@ module Api::Hses
   def host
     Rails.configuration.x.hses.api_hostname
   end
+
+  def request
+    Rails.logger.debug <<~EODM
+      HSES #{path}:
+      #{response.inspect}
+    EODM
+    if response.failed?
+      Rails.logger.error "HSES call to #{path} responded with #{response.code}"
+    end
+    response
+  end
 end
 
 class Api::Hses::Issue < ApiRequest
   include Api::Hses
-  include FakeApiResponseWrapper
   attr_accessor :id
 
   def initialize(id:)
     @id = id
   end
 
-  # TODO update this with real data once we have /issue endpoint
-  def request
-    fake_issue = FakeIssues.instance.data.find { |c| c[:id] == id }
-    details_response(fake_issue || FakeIssues.instance.data.last)
+  private
+
+  def path
+    "/issues-ws/issue/#{id}"
+  end
+
+  def parameters
+    {}
   end
 end
 
@@ -42,17 +56,6 @@ class Api::Hses::Issues < ApiRequest
   def initialize(user:, params: {})
     @username = user["uid"]
     @page = params[:page] || 1
-  end
-
-  def request
-    Rails.logger.debug <<~EODM
-      HSES #{path}:
-      #{JSON.pretty_generate(response)}
-    EODM
-    if response.failed?
-      Rails.logger.error "HSES call to #{path} responded with #{response.code}"
-    end
-    response
   end
 
   def response_type
--- a/spec/models/api/hses_spec.rb
+++ b/spec/models/api/hses_spec.rb
@@ -19,9 +19,23 @@ RSpec.describe Api::Hses do
     end
 
     describe "#request" do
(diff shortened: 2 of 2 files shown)


app/models/api/hses.rb
```diff
@@ -16,21 +16,35 @@ def error_type
...
-    fake_issue = FakeIssues.instance.data.find { |c| c[:id] == id }
-    details_response(fake_issue || FakeIssues.instance.data.last)
+  private
+
+  def path
+    "/issues-ws/issue/#{id}"
+  end
+
+  def parameters
+    {types: 1}
```
jduss4, line 47: [question] Can you explain why you are sending `types` as part of the URL? I don't see in the spec where we would be sending additional parameters for a specific issue besides the issue id in the path.

Sgtpluck: i dunno, i thought it was needed! but i guess it's not so i will take it out

app/models/api/hses.rb
```diff
@@ -16,21 +16,35 @@ def error_type
   def host
     Rails.configuration.x.hses.api_hostname
   end
+
+  def request
```
rahearn, line 20: [praise] oh that's awesome that we can move this whole method up into the parent module

app/models/api/hses.rb
```diff
@@ -16,21 +16,35 @@ def error_type
...
   end
 
-  # TODO update this with real data once we have /issue endpoint
-  def request
-    fake_issue = FakeIssues.instance.data.find { |c| c[:id] == id }
-    details_response(fake_issue || FakeIssues.instance.data.last)
+  private
+
+  def path
+    "/issues-ws/issue/#{id}"
```
rahearn, line 43: [thought] should we be using some sort of URL builder here instead of string interpolation? I'm trying to think of some sort of `id` that would result in really weird behavior, and/or a security hole.

rahearn (approved):
That `id` is coming from the complaints list API responses, so I'm thinking that the answer to my question is "no"

app/models/api/hses.rb
Sgtpluck: actually we could move it to the ApiResponse object if we really wanted to (i didn't do it yet cuz i wanted to keep it small) and remove it from TTA

app/models/api/hses.rb
jduss4: Hmm.  You want like, `File.join("directory", "file")` but for URLs?  This whole thing ultimately gets run through `URI::HTTPS.build(uri_params)` but I guess I don't know what kind of safeguards that has.

app/models/api/hses.rb
rahearn: (assuming you mean `ApiRequest`) I think that'd be a good thing to keep in our back pocket and do if/when the pattern is the same for monitoring as well.

app/models/api/hses.rb
Sgtpluck: i think it has to be an integer, so we could add a check -- in terms of a builder, can you tell me more about what you might be envisioning? this is just copying what the other similar path methods are doing, so i didn't think about it much

rahearn: > This whole thing ultimately gets run through `URI::HTTPS.build(uri_params)`

good point! Some quick testing reveals that the most obvious weird thing to include, a space, results in an `URI::InvalidComponentError` so it's at least not blindly passing input through. This is probably fine

rahearn: > can you tell me more about what you might be envisioning

Jess nailed what I was thinking, and also why it's not needed

jduss4 (approved):

kbighorse (approved):

PR #216, opened 2021-10-30
Improve style on grantee details view

kbighorse:
## Description of change
- top of list now aligns with sidebar
- removed dividers in sidebar sections
- added vertical spacing between complaint cards

## Issue(s)

* closes #0

## Checklist

- [ ] Meets issue criteria
- [ ] Boundary diagram updated

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/assets/stylesheets/complaints.scss
+++ b/app/assets/stylesheets/complaints.scss
@@ -2,6 +2,7 @@
   background-color: #fff;
   border: 2px solid #dfe1e2;
   border-radius: 5px;
+  padding-bottom: 20px;
 }
 
 .summary-header {
@@ -16,6 +17,7 @@
 
 .summary-body {
   padding: 0 1rem;
+  padding-bottom: 1rem;
 }
 
 .usa-process-list__item.usa-process-list__item-disc::before {
--- a/app/views/grantees/show.html.erb
+++ b/app/views/grantees/show.html.erb
@@ -8,30 +8,26 @@
   <div class="grid-row grid-gap">
     <div class="grid-col-4">
       <div class="summary-section summary-body">
-        <div class="ct-section-divider">
-          <h2>Summary</h2>
-          <p class="display-flex flex-justify">
-            <span> Total Centers:</span> 
-            <span><%= @grantee.centers_total %> </span>
-          </p>
-          <p class="display-flex flex-justify">
-            <span> FY21 Complaints:</span> 
-            <span><%= @grantee.complaints_per_fy %> </span>
-          </p>
-          <a href="<%= @grantee.hses_link %>" 
-          class="usa-button width-full" target="_blank">Open in HSES</a>
-        </div>
+        <h2>Summary</h2>
+        <p class="display-flex flex-justify">
+          <span> Total Centers:</span> 
+          <span><%= @grantee.centers_total %> </span>
+        </p>
+        <p class="display-flex flex-justify">
+          <span> FY21 Complaints:</span> 
+          <span><%= @grantee.complaints_per_fy %> </span>
+        </p>
+        <a href="<%= @grantee.hses_link %>" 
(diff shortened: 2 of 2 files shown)


kbighorse:
image

kbighorse:
currently stuck on adding vertical spacing between sidebar sections. also possibly mergeable now.

app/views/grantees/show.html.erb
```diff
@@ -8,30 +8,26 @@
...
       <div class="summary-section summary-body">
-        <div class="ct-section-divider">
-          <h2>Open Issues</h2>
-        </div>
+        <h2>Open Issues</h2>
       </div>
     </div>
     <div class="grid-col-8">
-      <ul class="usa-process-list">
+      <ul class="usa-process-list" style="padding-top: 0rem;">
```
rahearn, line 28: [suggestion] I think putting inline styles into here is a bit of a code smell. Either we should be able to customize the class, or add a new one, or otherwise fix things up rather than overriding the padding so directly.

app/assets/stylesheets/complaints.scss
```diff
@@ -2,6 +2,7 @@
   background-color: #fff;
   border: 2px solid #dfe1e2;
   border-radius: 5px;
+  padding-bottom: 20px;
```
rahearn, line 5: [suggestion] I think we've been mostly keeping spacing using the `rem` unit and it would be good to continue that.

rahearn (changes requested):
I really like getting the extra space between the complaint cards, but I think we can cleanup the implementation to not use inline styles

app/views/grantees/show.html.erb
kbighorse: I tried `usa-process-list` but I think something up the chain overrode it and then `ul` felt weird. Maybe a new class is the answer, but feels like overkill too.

app/assets/stylesheets/complaints.scss
jduss4: I added something in a recent commit that may be of help, or it might just complicate matters depending on what you're aiming for:

https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/blob/main/app/assets/stylesheets/complaints.scss#L5

app/views/grantees/show.html.erb
jduss4: [suggestion] My preference is also not to have inline styles. Could you use something like this USWDS class?
`.padding-bottom-1`

https://designsystem.digital.gov/utilities/margin-and-padding/

app/views/grantees/show.html.erb
```diff
@@ -8,30 +8,26 @@
...
         <% @grantee.complaints.each do |complaint| %>
-          <li>
+          <li style="padding-bottom: 20px;">
```
jduss4, line 30: [suggestion] Same idea here, maybe you can rely on an USWDS padding-bottom class instead of an inline style?

jduss4: I realize my other comment got a little lost in a resolved conversation, so here it is for easier viewing:

[suggestion] My preference is also not to have inline styles. Could you use something like this USWDS class?
.padding-bottom-1

https://designsystem.digital.gov/utilities/margin-and-padding/

Gemfile.lock
```diff
@@ -66,15 +66,14 @@ GEM
     bindex (0.8.1)
     bootsnap (1.9.1)
       msgpack (~> 1.0)
-    brakeman (5.1.2)
+    brakeman (5.1.1)
```
jduss4, line 69: [concern] I think that something went awry with a merge, because it looks like a lot of changes from the main branch are getting walked back in this commit!

jduss4 (changes requested):

Issue #221, opened 2021-11-01
Task for /complaints list: YES for RAN or TTA


jduss4:
If a particular complaint has a record in the database for monitoring or TTA activity, you can assume that there has been an association made, so add "YES" to the column on the /complaints page.

Originally @lauranash18f hoped to have more information to display to the user:

![image](https://user-images.githubusercontent.com/2480492/139736349-df76566e-e3e7-4aab-985b-b2a9bdd9c19b.png)

Ignore the numbers, No, TBD, etc. We don't have enough info to fill that out.

PR #223, opened 2021-11-02
changes table detail to use complaint model and adds RAN / TTA activity

jduss4:
## Description of change

I went to add "Yes" to the RAN and TTA columns of the index page's complaint table, when I decided it would be much easier if the Complaint model itself knew whether it had any RAN / TTA activity.  Then I realized heck, it would be great if it knew if it was a new complaint, and if it could return its status already formatted and then I kinda just kept redoing things....

- moves formatted status behavior out of helpers, simplifies views
- adds methods for `new?`, `has_monitoring_review?`, and `has_tta_report?`

In terms of the RAN / TTA issue, this is what I've got:

image

## Acceptance Criteria

- are these helpful and logical changes?
- nothing got broken along the way (like status formatting / sorting)
- RAN and TTA have "Yes" if the issue is linked to something

## How to test

Start the app, navigate to a complaint page and add a TTA and / or RAN link. Now on the complaints page, you should see "Yes" in the appropriate column.

## Issue(s)

* closes https://github.com/OHS-Hosting-Infrastructure/complaint-tracker/issues/221

## Checklist

- [X] Meets issue criteria
- [X] Code is meaningfully tested
- [X] Meets accessibility standards (WCAG 2.1 Levels A, AA)

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/models/complaint.rb
+++ b/app/models/complaint.rb
@@ -9,10 +9,24 @@ class Complaint
     initialContactDate: "Initial contact from complaint"
   }.with_indifferent_access.freeze
 
+  FORMATTED_STATUS = {
+    0 => "In Progress",
+    1 => "Closed",
+    2 => "Rec. Closure",
+    3 => "Rec. Reopening",
+    4 => "Reopened"
+  }.with_indifferent_access.freeze
+
   def initialize(hses_complaint)
     @id = hses_complaint[:id]
     @attributes = hses_complaint[:attributes].with_indifferent_access
     @links = hses_complaint[:links].with_indifferent_access
+  end
+
+  def creation_date
+    Date.parse(attributes[:creationDate])
+  rescue
+    nil
   end
 
   def formatted_creation_date
@@ -35,6 +49,14 @@ class Complaint
     attributes[:grantee]
   end
 
+  def has_monitoring_review?
+    @has_monitoring_review ||= IssueMonitoringReview.where(issue_id: id).any?
+  end
+
+  def has_tta_report?
+    @has_tta_report ||= IssueTtaReport.where(issue_id: id).any?
+  end
+
   def overdue?
     due_date? && due_date.before?(Date.current)
   end
@@ -55,12 +77,18 @@ class Complaint
     links[:html]
   end
 
+  def new?
+    # status_id 0 maps to "Open" label in HSES response
+    creation_date > 1.week.ago && status_id == 0
+  end
+
   def priority
     attributes[:priority][:label]
   end
 
   def status
-    attributes[:status][:label]
+    # we are changing to our own labels rather than the ones in HSES
+    new? ? "New" : FORMATTED_STATUS[status_id]
   end
 
   def summary
@@ -83,12 +111,6 @@ class Complaint
     due_date&.strftime("%m/%d/%Y")
   end
 
-  def creation_date
-    Date.parse(attributes[:creationDate])
-  rescue
-    nil
-  end
-
   def issue_last_updated
     Date.parse(attributes[:issueLastUpdated])
   rescue
@@ -97,5 +119,9 @@ class Complaint
 
   def relative_time_til_due
     (Date.current...due_date).count
+  end
+
+  def status_id
+    attributes[:status][:id]
   end
 end
--- a/app/models/complaint.rb
+++ b/app/models/complaint.rb
@@ -15,7 +15,7 @@ class Complaint
     2 => "Rec. Closure",
     3 => "Rec. Reopening",
     4 => "Reopened"
-  }.with_indifferent_access.freeze
+  }.freeze
 
   def initialize(hses_complaint)
     @id = hses_complaint[:id]
--- a/app/views/complaints/_complaints_table.html.erb
+++ b/app/views/complaints/_complaints_table.html.erb
@@ -16,7 +16,7 @@
   </thead>
   <tbody>
   <% @complaints.each do |complaint| %>
-    <%= render "complaints/complaints_table_detail", complaint: complaint %>
+    <%= render "complaints/complaints_table_detail", complaint_data: complaint %>
   <% end %>
   </tbody>
 </table>
--- a/app/views/complaints/_complaints_table_detail.html.erb
+++ b/app/views/complaints/_complaints_table_detail.html.erb
@@ -1,22 +1,23 @@
+<% complaint = Complaint.new(complaint_data) %>
+
 <tr>
-  <% formatted_status = status(complaint[:attributes]) %>
   <th scope="row" role="rowheader">
-    <span class="<%= formatted_status == "New" ? "text-bold" : "" %>">
-      <%= link_to complaint[:id], complaint_path(complaint[:id]), class: "usa-link" %>
+    <span class="<%= complaint.new? ? "text-bold" : "" %>">
+      <%= link_to complaint.id, complaint_path(complaint.id), class: "usa-link" %>
     </span>
   </th>
   <td
-    data-sort-value="<%= DateTime.parse(complaint[:attributes][:creationDate]).to_i %>"
+    data-sort-value="<%= sort_date(complaint.creation_date) %>"
   >
-    <%= DateTime.parse(complaint[:attributes][:creationDate]).strftime("%m/%d/%Y") %>
+    <%= complaint.formatted_creation_date %>
   </td>
   <td
-    data-sort-value="<%= ComplaintsHelper::STATUS_SORT_ORDER[formatted_status] %>"
+    data-sort-value="<%= sort_status(complaint.status) %>"
   >
-    <%= formatted_status %>
+    <%= complaint.status %>
   </td>
-  <td><%= complaint[:attributes][:grantee] %></td>
-  <td><%= truncate(complaint[:attributes][:summary]) %></td>
-  <td> TBD </td>
-  <td> TBD </td>
+  <td><%= complaint.grantee %></td>
+  <td><%= truncate(complaint.summary) %></td>
+  <td><%= "Yes" if complaint.has_tta_report? %></td>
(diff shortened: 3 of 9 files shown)


app/models/complaint.rb
```diff
@@ -9,12 +9,26 @@ class Complaint
     initialContactDate: "Initial contact from complaint"
   }.with_indifferent_access.freeze
 
+  FORMATTED_STATUS = {
+    0 => "In Progress",
+    1 => "Closed",
+    2 => "Rec. Closure",
+    3 => "Rec. Reopening",
+    4 => "Reopened"
+  }.with_indifferent_access.freeze
```
jduss4, line 18: [question] in retrospect, maybe indifferent access doesn't matter here because I don't think numbers can be symbols...?

app/views/complaints/_complaints_table.html.erb
```diff
@@ -16,7 +16,7 @@
   </thead>
   <tbody>
   <% @complaints.each do |complaint| %>
-    <%= render "complaints/complaints_table_detail", complaint: complaint %>
+    <%= render "complaints/complaints_table_detail", complaint_data: complaint %>
```
jduss4, line 19: [question] I decided to pass in the data and stick it in `Complaint.new` in the partial, so that it was more clear in the partial what sort of data it was dealing with, but I think either way is probably fine. Any preferences?

Sgtpluck: [thought] i wonder if it makes more sense to take a step up the chain and make `@complaints` a collection of complaint objects. i think that would be my preference. that and making this a collection partial, if possible (i know there are some funky naming conventions that can make that tricky)

app/views/complaints/_complaints_table_detail.html.erb
```diff
@@ -1,22 +1,23 @@
...
   <th scope="row" role="rowheader">
-    <span class="<%= formatted_status == "New" ? "text-bold" : "" %>">
-      <%= link_to complaint[:id], complaint_path(complaint[:id]), class: "usa-link" %>
+    <span class="<%= complaint.new? ? "text-bold" : "" %>">
+      <%= link_to complaint.id, complaint_path(complaint.id), class: "usa-link" %>
     </span>
   </th>
   <td
-    data-sort-value="<%= DateTime.parse(complaint[:attributes][:creationDate]).to_i %>"
+    data-sort-value="<%= sort_date(complaint.creation_date) %>"
```
Sgtpluck, line 10: [thought] i'm thinking about whether there'd be drawbacks to just putting sort_date and sort_status on the complaint object. it's view-centric, so i get why it's in a helper. this is out of scope, but it would be a good opportunity for a view decorator!

app/views/complaints/_complaints_table_detail.html.erb
```diff
@@ -1,22 +1,23 @@
+<% complaint = Complaint.new(complaint_data) %>
+
 <tr>
-  <% formatted_status = status(complaint[:attributes]) %>
   <th scope="row" role="rowheader">
-    <span class="<%= formatted_status == "New" ? "text-bold" : "" %>">
-      <%= link_to complaint[:id], complaint_path(complaint[:id]), class: "usa-link" %>
+    <span class="<%= complaint.new? ? "text-bold" : "" %>">
```
Sgtpluck, line 5: [praise] i LOVE these new methods

app/views/complaints/_complaints_table_detail.html.erb
```diff
@@ -1,22 +1,23 @@
...
-    <%= formatted_status %>
+    <%= complaint.status %>
   </td>
-  <td><%= complaint[:attributes][:grantee] %></td>
-  <td><%= truncate(complaint[:attributes][:summary]) %></td>
-  <td> TBD </td>
-  <td> TBD </td>
+  <td><%= complaint.grantee %></td>
+  <td><%= truncate(complaint.summary) %></td>
+  <td><%= "Yes" if complaint.has_tta_report? %></td>
```
Sgtpluck, line 21: [question] should it have No if it doesn't have a report?

spec/models/complaint_spec.rb
```diff
@@ -53,6 +79,71 @@
...
+        expect(subject.has_tta_report?).to be false
+      end
+    end
+  end
+
+  describe "#new?" do
+    context "complaint is less than a week old" do
+      context "an open complaint" do
+        before { subject.attributes[:status][:id] = 0 }
+        it "is new" do
```
Sgtpluck, line 116: [nit] pls add a newline after the before block

spec/models/complaint_spec.rb
```diff
@@ -53,6 +79,71 @@
...
+        it "is not new" do
+          expect(subject.new?).to be false
+        end
+      end
+    end
+
+    context "complaint is more than a week old" do
+      before { subject.attributes[:creationDate] = 8.days.before.to_date.iso8601 }
+      before { subject.attributes[:status][:id] = 0 }
+      context "an open complaint" do
```
Sgtpluck, line 132: [nit] please add a newline before `context` block

spec/models/complaint_spec.rb
```diff
@@ -146,6 +237,38 @@
...
+      before { subject.attributes[:status][:id] = 0 }
+      it "returns 'New'" do
+        expect(subject.status).to eq "New"
+      end
+    end
+
+    context "a complaint that is more than a week old and open" do
+      before { subject.attributes[:creationDate] = 8.days.before.to_date.iso8601 }
+      before { subject.attributes[:status][:id] = 0 }
+      it "returns 'In Progress'" do
```
Sgtpluck, line 252: [nit] please add a newline before it block

Sgtpluck (changes requested):
it looks dope. i had some thoughts but i love it

app/views/complaints/_complaints_table.html.erb
jduss4: I thought about that and then decided not to pull on the thread. I mean, presumably that could be happening at the api.request sort of level but I didn't want to have to change things around alllll over the app when I was ostensibly just here to add a Yes for RAN / TTA, lol.  I do think it's a good idea, though.  What do you think?  This PR or should I make an issue for a future one?

app/models/complaint.rb
```diff
@@ -9,12 +9,26 @@ class Complaint
...
+
   def initialize(hses_complaint)
     @id = hses_complaint[:id]
     @attributes = hses_complaint[:attributes].with_indifferent_access
     @links = hses_complaint[:links].with_indifferent_access
   end
 
+  def creation_date
```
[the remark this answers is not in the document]
jduss4: Seems possible, although I don't know as that we're duplicating it. They all pull different field names and complaint has slightly different behavior (rescuing). It wouldn't be hard to put something into a concern and pass the field name in, though?

app/views/complaints/_complaints_table_detail.html.erb
jduss4: Me too!!!!! 😄

app/views/complaints/_complaints_table_detail.html.erb
jduss4: I don't know much about view decorators to weigh in on that aspect, but it definitely seems like there's probably a better approach to this

app/views/complaints/_complaints_table_detail.html.erb
jduss4: @lauranash18f and I decided not to have anything about that, because if there isn't an activity report, say, for a complaint, then we have no way of actually knowing it doesn't exist vs only knowing that nothing has been linked.  In the original design, it had No / TBD / Yes but realistically all we can mark is yes unless if we simply mean that "no" means "nothing linked as far as the complaint tracker is concerned"

app/views/complaints/_complaints_table.html.erb
rahearn: I think it probably makes the most sense to have the Api::Hses class wrap the response in the complaint objects. I'd vote for doing that here, but I don't mind if it's done in a future one either.

app/models/complaint.rb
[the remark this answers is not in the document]
rahearn: I'd vote for not pulling anything out if you then have to pass extra arguments in to make use of it.

app/views/complaints/_complaints_table_detail.html.erb
rahearn: ➕

rahearn (commented):
Approved if it's decided to wrap the complaint data in a separate PR.

app/views/complaints/_complaints_table.html.erb
jduss4: Okay. I'll do that here, but leave the other `Api::x` classes alone for now!

jduss4: HMM actually upon further review, I'm having a little trouble figuring out how would be best to go about doing this. Can we merge as is and then open a ticket for thinking about where and how to do it within an Api::Hses type location, or in ApiResponse, etc?  I kinda think it's going to be better to do everything at once rather than just complaints.

app/models/complaint.rb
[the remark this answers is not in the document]
Sgtpluck: yeah seems like another thing that would be good for a decorator, but not worth it if we can't use it directly

app/views/complaints/_complaints_table.html.erb
Sgtpluck: oh i was just thinking we'd map them in the controller, since we're using the object in the view.

Sgtpluck (approved):

PR #230, opened 2021-11-05
Add sort variable to pagination links

### LINKED ISSUE #229: Paging through complaints resets the sort back to default
**Describe the bug**

The paging links don't preserve the current sort setting.

**To Reproduce**
1. Sort complaints by oldest first
2. Click on page 2
3. You'll see the second page of most recent complaints

**Expected behavior**
Step 3 should have been the second page of oldest complaints

rahearn:
## Description of change

Adds sort to the pagination links so we can access the second page of oldest issues

## Acceptance Criteria

Sort issues by oldest first
Click on link to page 2
Should see second-oldest page of issues

## Issue(s)

* closes #229 

## Checklist

- [x] Meets issue criteria

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/views/complaints/_complaints_table.html.erb
+++ b/app/views/complaints/_complaints_table.html.erb
@@ -29,6 +29,6 @@
   <!-- this content will change when sort changes -->
 </div>
 
-<%= render partial: 'complaints/pagination', locals: {pagy: @pagy} %>
+<%= render partial: 'complaints/pagination', locals: {pagy: @pagy, path_args: {sort: @sort}} %>
 
 <%= javascript_pack_tag 'complaint-sort' %>
--- a/spec/views/complaints/_pagination.erb_spec.rb
+++ b/spec/views/complaints/_pagination.erb_spec.rb
@@ -10,6 +10,20 @@ RSpec.describe "rendering pagination" do
       expect(rendered).to include '<a class="usa-pagination__button usa-current" aria-label="Page 1" aria-current="page" href="/complaints?page=1">1</a>'
     end
 
+    it "generates proper link for previous page" do
+      @pagy = Pagy.new(count: 250, page: 2)
+
+      render partial: "complaints/pagination", locals: {pagy: @pagy, path_args: {sort: "creationDate"}}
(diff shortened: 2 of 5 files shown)


app/views/complaints/_complaints_table.html.erb
```diff
@@ -29,6 +29,6 @@
   <!-- this content will change when sort changes -->
 </div>
 
-<%= render partial: 'complaints/pagination', locals: {pagy: @pagy} %>
+<%= render partial: 'complaints/pagination', locals: {pagy: @pagy, path_args: {sort: @sort}} %>
```
jduss4, line 32: [thought] I wonder if eventually, assuming we had filters, etc, if we would want to use a clone of `params` rather than listing individual parameters. idk. For now I think this is fine, just thinking out loud.

spec/views/complaints/_pagination.erb_spec.rb
```diff
@@ -22,7 +22,7 @@
     it "has no pagination" do
       @pagy = Pagy.new(count: 25)
 
-      render partial: "complaints/pagination", locals: {pagy: @pagy}
+      render partial: "complaints/pagination", locals: {pagy: @pagy, path_args: {}}
```
jduss4, line 25: [suggestion] I realize it's being tested above, but I don't hate the idea of explicitly testing that arguments are being treated correctly by this partial, too.  Thoughts?

jduss4 (approved):
A couple non-blocking questions, but it looks good to me!

app/views/complaints/_complaints_table.html.erb
rahearn: Might be worthwhile later, but I think for now being explicit is better.

spec/views/complaints/_pagination.erb_spec.rb
rahearn: Hmm, I'll add checks for previous and next, since they're generated in the partial and not the helper.

PR #238, opened 2021-11-10
Add validation for Display Id format

Sgtpluck:
## Description of change
Adds some validation around the TTA Display ID format because TTAHub 500s if the display id is not in the expected format.

## Acceptance Criteria
When I input a TTA Activity Report Display ID that does not match the format `/^R/d{2}-AR-/`
I want an error that tells me it's a bad format
So I know what the problem is

## How to test

- [ ] Go to a complaint detail page.
- [ ] Try to link a TTA AR ID that is not in the `/^R/d{2}-AR-/` format
- [ ] You should get an error that says "This doesn't look like a TTA Activity Report Display ID. Please double check the format."
- [ ] Try to like a TTA AR ID that is in that format! You should get a different response depending on whether it's an ID you have access to or not.

## Checklist

## To the Reviewer

* **praise:**
* **nitpick:**
* **suggestion:**
* **issue:**
* **question:**
* **thought:**
* **chore:**

### DIFF
--- a/app/models/tta_activity_report.rb
+++ b/app/models/tta_activity_report.rb
@@ -2,6 +2,7 @@ require "api_delegator"
 
 class TtaActivityReport
   include ActiveModel::Validations
+  validate :check_display_id_format
   validate :api_call_succeeded
 
   attr_accessor :access_token, :display_id
@@ -38,6 +39,7 @@ class TtaActivityReport
   private
 
   def api_call_succeeded
+    return unless errors.blank?
     errors.add(:base, tta_link_error_message) if activity_data.failed?
   end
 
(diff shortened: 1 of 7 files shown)


app/models/tta_activity_report.rb
```diff
@@ -38,6 +39,7 @@ def api_error_message
   private
 
   def api_call_succeeded
+    return unless errors.blank?
```
Sgtpluck, line 42: this way, the API call won't happen if the display id is the wrong format

app/models/tta_activity_report.rb
```diff
@@ -57,6 +59,12 @@ def api
     ApiDelegator.use("tta", "activity_report", display_id: display_id, access_token: access_token)
   end
 
+  def check_display_id_format
+    unless display_id.match?(/^R\d{2}-AR-/)
```
jduss4, line 63: [comment, non-blocking] I can't find where I've seen the legacy display id format, but I think this regex will only work for the current format.  I think it's good enough for our current purposes, though.  Do you think it's worth adding a comment in case anybody is working from our code?

jduss4 (approved):

app/models/tta_activity_report.rb
Sgtpluck: sure

